PR #3, opened 2020-01-27
rewrite model test to unittest

jajcayn:
so now you can run individual test as
```bash
python tests/test_models.py
```
or when having multiple tests you can just
```bash
pytest tests
```


jajcayn:
I also switch travis CI to `pytest` si when we add tests, they are automatically run

PR #7, opened 2020-01-27
Formatting

jajcayn:
* reformated all files with black, BUT kept the line wrapping
* if you'd like to have the same intermediate setting as myself, do `black --line-length 1000`
* sorted imports using `isort`


jajcayn:
I am fine... we can merge this and then #14 has to be rebased on the current master (i.e. after merging this)...

PR #13, opened 2020-01-27
Helpers for connectivity

jajcayn:
Useful functions and methods working on connectivities and atlases.
* 4 types: structural, functional, fiber lengths and delays
* common functionality:
    * loading from `mat` or `h5` file, initializing single node dummy connectivity
    * indexing upper/lower triangle for correlation measures
    * functions for adjusting density, symmetrizing, nullifying diagonal and others
    * export to networkx's `Graph`
* specific functionality:
    * structural matrices: scaling to relative (0-1 range)
    * functional matrices: initializing from timeseries by computing correlation or covariance matrices
    * fiber length matrices: compute delay matrices using signal speed
    * delay matrices: max delay, putting a unit delay on diagonal, transforming delay matrix into multiples of dts for actual integration

Helpers for an ensemble of connectivities:
* initialize from folder - i.e. read all files
* fast computation of mean and median matrices

Also, I wrote some tests for it.

Finally, some little Santa's helper for Atlases - so far only AAL2 and DK, these atlases mainly just state node numbers with their names, but I suggest in future they could e.g. output means per lobes, so if your timeseries data from model is AAL2, you could do frontal lobe average easily as `np.mean(data[sc_matrix.atlas.get_frontal_nodes, :], axis=0)`


caglorithm:
Great, I'll have a look later! 
The files seem very helpful, especially the atlases! (Impressive collection :)

Concerning the upper part: I don't think it's necessary to put a `(c)` up there, there is a licence in the project which covers the whole repository. People often do it, though it seems to me more like an alibi thing. We could put simply "neurolib" there, in case anyone wonders where this file comes from :D

A description of the file seems helpful though, since it's not always clear what the purpose of the file is to a user.

tests/models/test_models.py
```diff
@@ -12,33 +10,19 @@ class TestALNModel(unittest.TestCase):
     """
 
     def test_single_node(self):
-        logging.info("\t > ALN: Testing single node ...")
-        start = time.time()
-
```
caglorithm: No output and measurements for testing anymore?

jajcayn: when testing with a suite (`unittest` or `pytest`), you don't need to: you can't see the output anyway and testing is timing everything by default, you can later check the results so no need to do it explicitely within the testing script

jajcayn: so e.g. `pytest --durations=0 tests/` shows:
```
=========================================================================================== test session starts ============================================================================================
platform darwin -- Python 3.7.6, pytest-5.3.4, py-1.8.1, pluggy-0.13.1
rootdir: /Users/nikola/work-brain/neurolib
collected 19 items

tests/models/test_models.py ....                                                                                                                                                                     [ 21%]
tests/utils/test_connectivity.py ...............                                                                                                                                                     [100%]

============================================================================================= warnings summary =============================================================================================
tests/models/test_models.py::TestALNModel::test_network
tests/models/test_models.py::TestALNModel::test_network
....
-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================================== slowest test durations ==========================================================================================
15.41s call     tests/models/test_models.py::TestALNModel::test_network
6.02s call     tests/models/test_models.py::TestHopfModel::test_network
3.09s call     tests/models/test_models.py::TestALNModel::test_single_node
0.68s call     tests/models/test_models.py::TestHopfModel::test_single_node
0.03s call     tests/utils/test_connectivity.py::TestFunctionalConnectivity::test_init_from_timeseries
0.01s call     tests/utils/test_connectivity.py::TestBaseConnectivity::test_load_from_file

(0.00 durations hidden.  Use -vv to show these durations.)
```

caglorithm: Perfect!

requirements.txt
```diff
@@ -4,3 +4,4 @@ numpy
 scipy
 pytest
 flake8
+networkx
```
caglorithm: I think this is a considerable requirement!

jajcayn: sure, agreed, no need to add it, I can just delete the function that does `Graph` out of it... if anyone needs it, they can do it afterwards by themselves!

jajcayn:
so as per our discussion, closing now, keeping the branch; we will focus on modelling functionality for now and then work on helpers like this..

PR #29, opened 2020-02-06
Feature/signal class

jajcayn:
for now just the class itself, please have a look and possibly comment on capabilities; i.e. DO NOT MERGE now please, just after I add tests

**to be added:**
* [x] tests for `Signal` class
* [x] small notebook showcasing the `Signal` capabilities


neurolib/utils/signal.py
```diff
@@ -0,0 +1,515 @@
...
+        "name",
+        "label",
+        "signal_type",
+        "unit",
+        "description",
+        "process_steps",
+    ]
+
+    @classmethod
+    def from_file(cls, filename):
```
caglorithm: What's the use case scenario of this? Is this so the user can save and load processed signals to disk? Should we have this for model outputs as well?

neurolib/utils/signal.py
```diff
@@ -0,0 +1,515 @@
...
+        """
+        Load signal from saved file.
+
+        :param filename: filename for the Signal
+        :type filename: str
+        """
+        if not filename.endswith(NC_EXT):
+            filename += NC_EXT
+        # load NC file
+        xarray = xr.load_dataarray(filename)
```
caglorithm: Maybe we should adopt a unified way of saving arrays / signals / etc. This sounds interesting, if we adopt `xarray` throughout the project.

neurolib/utils/signal.py
```diff
@@ -0,0 +1,515 @@
...
+Base classes for representing signals.
+"""
+
+import logging
+from copy import deepcopy
+
+import numpy as np
+import xarray as xr
+from mne import set_log_level
+from mne.filter import filter_data, resample
```
caglorithm: If you could do this without mne but with scipy.filter (which we already require) you'd be the mvp!

requirements.txt
```diff
@@ -8,3 +8,4 @@ pandas
 tqdm
 pypet
 deap
+mne
```
caglorithm: 💻🔥

neurolib/utils/signal.py
jajcayn: typically you'd do something like

```python
signal = RatesSignal.from_file("filename_of_saved_data.nc")
```

which is shorter and more concise than something like:
```
array = xr.load_dataarray("filename_of_saved_data.nc")
signal = RatesSignal(data=array)
```

caglorithm:
LGTM! Do you wanna merge after making the notebook or before?

jajcayn:
> LGTM! Do you wanna merge after making the notebook or before?

after pls, I am in the process of making it, just be approx 30 more mins)

caglorithm:
> > LGTM! Do you wanna merge after making the notebook or before?
> 
> after pls, I am in the process of making it, just be approx 30 more mins)

Take your time pls, no hurry :)

jajcayn:
@caglorithm all done, after the tests run and you check the notebook, we can merge :)

caglorithm:
Congrats on a very nice and extensive notebook! I just found some minor errors here and there but commenting on a notebook is not easy here on github so let's just accept it and correct it later! 

Merging!

PR #30, opened 2020-02-06
New testing cases

caglorithm:
Removed unnecessary functions and added a few new testing cases for evolution utility functions and for datasets. Matplotlib lazy importing now.

### DIFF
--- a/neurolib/optimize/evolution/evolutionaryUtils.py
+++ b/neurolib/optimize/evolution/evolutionaryUtils.py
@@ -98,7 +98,7 @@ def plotScoresDistribution(scores, gIdx, save_plots=None):
     plt.xlabel("Score")
     plt.ylabel("Count")
     if save_plots is not None:
-        print("Saving plot to {}".format(os.path.join(paths.FIGURES_DIR, "%s_hist_%i.png" % (save_plots, gIdx))))
+        logging.info("Saving plot to {}".format(os.path.join(paths.FIGURES_DIR, "%s_hist_%i.png" % (save_plots, gIdx))))
         plt.savefig(os.path.join(paths.FIGURES_DIR, "%s_hist_%i.png" % (save_plots, gIdx)))
     plt.show()
 
--- a/neurolib/optimize/evolution/evolutionaryUtils.py
+++ b/neurolib/optimize/evolution/evolutionaryUtils.py
@@ -57,7 +57,7 @@ def printParamDist(pop=None, paramInterval=None, gIdx=None):
     if paramInterval == None:
         paramInterval = self.paramInterval
 
-    print("Parameters dictribution (Generation {}):".format(gIdx))
+    print("Parameter dictribution (Generation {}):".format(gIdx))
     for idx, k in enumerate(paramInterval._fields):
         print(
             "{}: \t mean: {:.4},\t std: {:.4}".format(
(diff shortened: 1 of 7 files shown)


neurolib/optimize/evolution/evolutionaryUtils.py
```diff
@@ -94,18 +90,23 @@ def printIndividuals(pop, paramInterval, stats=False):
...
+    import matplotlib.pyplot as plt
+
     plt.figure(figsize=(4, 2))
     plt.hist(scores, color="grey", edgecolor="black", linewidth=1.2)
     plt.title("Generation: %i, Individuals: %i" % (gIdx, len(scores)))
     plt.xlabel("Score")
     plt.ylabel("Count")
     if save_plots is not None:
-        logging.info("Saving plot to {}".format(os.path.join(paths.FIGURES_DIR, "%s_hist_%i.png" % (save_plots, gIdx))))
+        print("Saving plot to {}".format(os.path.join(paths.FIGURES_DIR, "%s_hist_%i.png" % (save_plots, gIdx))))
```
jajcayn: use f-string here

neurolib/utils/loadData.py
```diff
@@ -81,42 +81,56 @@ def loadData(
...
-                    print('\tLoaded key "{}"'.format(key))
-            elif type(matrix) is dict:
-                raise ValueError("Object is still a dict. Here are the keys: {}".format(matrix.keys()))
-            return matrix
-        except:  # Deco files
-            matrix = scipy.io.loadmat(matFileName)
+        matrix = scipy.io.loadmat(matFileName)
+        if verbose:
+            print("\tLoading using scipy.io.loadmat...")
+            print("Keys: {}".format(list(matrix.keys())))
```
jajcayn: f-string

neurolib/utils/loadData.py
```diff
@@ -81,42 +81,56 @@ def loadData(
             return matrices[0]
 
     def loadMatrix(self, matFileName, key="", verbose=False):
-        """
-        Function to load SC and FC .mat files of different formats.
+        """Function to furiously load SC and FC .mat files of different formats.
         """
         if verbose:
             print("Loading {}".format(matFileName))
```
jajcayn: f-string

neurolib/utils/loadData.py
```diff
@@ -81,42 +81,56 @@ def loadData(
...
-                print("\tLoading using scipy.io.loadmat...")
-                print("Keys: {}".format(list(matrix.keys())))
-            if key != "" and key in list(matrix.keys()):
-                matrix = matrix[key]
-                if verbose:
-                    print('\tLoaded key "{}"'.format(key))
-            elif type(matrix) is dict:
-                raise ValueError("Object is still a dict. Here are the keys: {}".format(matrix.keys()))
-            return matrix
+                print('\tLoaded key "{}"'.format(key))
```
jajcayn: f-string

tests/test_datasets.py
```diff
@@ -0,0 +1,31 @@
...
+import glob
+
+from neurolib.utils.loadData import Dataset
+import neurolib.utils.functions as func
+
+
+class TestDatasets(unittest.TestCase):
+    def test_dataset_gw(self):
+        ds = Dataset("gw")
+        assert ds.Cmat.shape == (80, 80)
```
jajcayn: not needed, but `unittest` provides very nice `self.assertTrue`, `assertEqual`, `assertListEqual`, etc.. better to use those since you will get much more clear messages if the test fails...
here you would use `self.assertTupleEqual`

similarly, all asserts below

neurolib/optimize/evolution/evolutionaryUtils.py
caglorithm: +1

tests/test_datasets.py
caglorithm: I can just replace all asserts? neat

jajcayn: yep, but it depends on what you are asserting... if general equality you can `self.assertEqual`, but you have a lot of options> https://docs.python.org/3/library/unittest.html#assert-methods

jajcayn: also, this is a nice cheat sheet: https://kapeli.com/cheat_sheets/Python_unittest_Assertions.docset/Contents/Resources/Documents/index

jajcayn:
LGTM , merging now

PR #39, opened 2020-02-21
Feature/autochunk

caglorithm:
I am excited about this (very big) PR which reaches an important milestone necessary to implement the upcoming `Circuit` class. Biggest change is that `chunkwiseIntegration.py` is now replaced by a model-agnostic chunkwise integrator in `model.py` called `autochunk`. It has the ability to simulate any model for one `dt` at a time and manage inputs and outputs of a model. In order to offer this ability for all possible models (including future ones), models can now register their variables, initial values, outputs and inputs and the rest is handled by the super class.
 
### Changes:
- Models now specify their `integration` function as an attribute. This function is called by the `Model` superclass for integration. `.run()` is now handled by the superclass instead of each model. Returned values of the `integration` function must be in the order `t, other_vars = integration()`.
- All models get rid of `chunkwiseIntegration.py`. A new `autochunk` method of the superclass `Model` handles chunkwise integration and BOLD simulation for all models, old and new.
- Super class `Model` has new attribute `.state` which captures the current state variables of the model. 
- For chunkwise integration, all models can specify the names of all variables `state_vars`, of all initial values `init_vars`, input variables `input_vars`, output variables `output_vars`, the default output variable `defaultOutput` (used for BOLD simulation) and the default input variable `defaultInput`. The last two can be used in the upcoming `Circuit` model for input-output mapping.
- As a preparation for the `Circuit` class, an `inputs` array can be passed to a model and be integrated a single `dt` using `model.run(onedt=True, inputs=inputs)`.
- For chunkwise integration, models can specify a `getMaxDelay` method, which is required for chunkwise integration with delays.
- `BOLDModel` is now initialized and called in the superclass `Model` instead of being handled by each model itself.
- The Balloon-Windkessel `BOLDModel` requires firing rate as an input (in `Hz`). Using attribute `normalize_bold_input = True`, models can now specify whether they output a firing rate (in `Hz`) or activity in some arbitrary unit. In the latter case, the activity is normalized automatically between `0` and `BOLDModel(normalize_max=50)`.
- Parameter dictionary `params` is now a `dotdict()` (from `neurolib.utils.collections`)
- 230 lines of code less.

### Todo:
- [x] Docstrings for new `Model` methods
- [x] `aln` model `ou` state variables should be called `mue_ou`
- [x] `Model` superclass requires each subclass to define an`integration` function as an attribute.
- [x] Chunkwise integration requires the model to report its state variables, initial values, outputs (for BOLD simulation). Refuses if the model reports them wrongly. Checks this at initialization.
- [x] If `getMaxDelay` is not given, it is either computed from the maximum value of the delay matrix `Dmat_ndt` or assumed to be `0`.
- [x] (Non-rate) models should be able to specify the maximum BOLD scaling (fixed at `50 Hz` now) 
- [x] Add BOLD functionality for normal integration (not chunkwise) 
- [x] In examples and tests, `bold=True` not necessary at model initialization. Instead use `run(bold=True)`.


neurolib-dev:
# [Codecov](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=h1) Report
> Merging [#39](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=desc) into [master](https://codecov.io/gh/neurolib-dev/neurolib/commit/f7030d1641000e7ce74b8ad87e6456467b81684a&el=desc) will **decrease** coverage by `0.83%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master      #39      +/-   ##
==========================================
- Coverage   95.40%   94.56%   -0.83%     
==========================================
  Files          28       25       -3     
  Lines        1803     1692     -111     
==========================================
- Hits         1720     1600     -120     
- Misses         83       92       +9     
```


| [Impacted Files](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [neurolib/utils/functions.py](https://codecov.io/gh/neurolib-dev/neurolib/pull/39/diff?src=pr&el=tree#diff-bmV1cm9saWIvdXRpbHMvZnVuY3Rpb25zLnB5) | `86.67% <0.00%> (-7.78%)` | :arrow_down: |
| [neurolib/models/model.py](https://codecov.io/gh/neurolib-dev/neurolib/pull/39/diff?src=pr&el=tree#diff-bmV1cm9saWIvbW9kZWxzL21vZGVsLnB5) | `92.44% <0.00%> (-5.21%)` | :arrow_down: |
| [neurolib/models/aln/loadDefaultParams.py](https://codecov.io/gh/neurolib-dev/neurolib/pull/39/diff?src=pr&el=tree#diff-bmV1cm9saWIvbW9kZWxzL2Fsbi9sb2FkRGVmYXVsdFBhcmFtcy5weQ==) | `99.15% <0.00%> (-0.12%)` | :arrow_down: |
| [neurolib/optimize/evolution/deapUtils.py](https://codecov.io/gh/neurolib-dev/neurolib/pull/39/diff?src=pr&el=tree#diff-bmV1cm9saWIvb3B0aW1pemUvZXZvbHV0aW9uL2RlYXBVdGlscy5weQ==) | `96.67% <0.00%> (+6.67%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=footer). Last update [f7030d1...35f8039](https://codecov.io/gh/neurolib-dev/neurolib/pull/39?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

neurolib/models/fhn/loadDefaultParams.py
```diff
@@ -38,7 +36,7 @@ class struct(object):
 
     # signal transmission speec between areas
     params.signalV = 20.0
-    params.K_gl = 250.0  # global coupling strength
+    params.K_gl = 0.6  # global coupling strength
```
jajcayn: 250 -> 0.6 is a pretty big difference 😄

neurolib/models/fhn/model.py
```diff
@@ -1,78 +1,55 @@
-import neurolib.models.fhn.chunkwiseIntegration as cw
+import numpy as np
```
jajcayn: not used

neurolib/models/aln/model.py
```diff
@@ -46,61 +77,35 @@ def __init__(
...
-        rates_inh = rates_inh * 1000.0
+    def getMaxDelay(self):
+        # compute maximum delay of model
+        dt = self.params["dt"]
+        Dmat = dp.computeDelayMatrix(self.params["lengthMat"], self.params["signalV"])
+        Dmat_ndt = np.around(Dmat / dt)  # delay matrix in multiples of dt
+        ndt_de = round(self.params["de"] / dt)
+        ndt_di = round(self.params["di"] / dt)
+        max_global_delay = int(max(np.amax(Dmat_ndt), ndt_de, ndt_di))
+        return max_global_delay
```
jajcayn: this is not much different from the superclass' implementation of `getMaxDelay` is it? you just also check `de` and `di` from params, right? so you could run the superclass' function and then just check whether local delays are not larger than `Dmat` max right? what about

```python
def getMaxDelay(self):
    ndt_de = round(self.params["de"] / dt)
    ndt_di = round(self.params["di"] / dt)
    max_dmat_delay = super().getMaxDelay()
    return int(max(max_dmat_delay, ndt_de, ndt_di))
```

neurolib/models/hopf/loadDefaultParams.py
```diff
@@ -1,8 +1,9 @@
 import numpy as np
-
 import h5py
```
jajcayn: both h5py and scipy are not used

neurolib/models/hopf/loadDefaultParams.py
```diff
@@ -58,21 +56,28 @@ class struct(object):
...
-    params_dict = params.__dict__
+    # values of the external inputs
+    params.x_ext = np.zeros((params.N,))
+    params.y_ext = np.zeros((params.N,))
 
-    return params_dict
+    return params
 
 
 def computeDelayMatrix(lengthMat, signalV, segmentLength=1):
```
jajcayn: this function is in all `loadDefaultParams.py` files (for each model). we might do some refactoring on this...
also, I would think about parameters, I personally do not like setting parameters inside python files... I am usually loading parameters from `yaml` or `json`... that way, we could provide `default_aln_params.yaml` for example and just do loading function...

neurolib/models/model.py
```diff
@@ -2,34 +2,321 @@
...
+        assert integration is not None, "Model integration function not given."
+        self.integration = integration
+
+        assert isinstance(params, dict), "Parameters must be a dictionary."
+        self.params = dotdict(params)
 
-    def __init__(self, name, description=None):
-        assert isinstance(name, str), f"Model name is not a string."
-        self.name = name
+        self.state_vars = state_vars
```
jajcayn: as I was saying, when in child class you are setting this outside the scope of `__init__` as attributes, you do not have to assign them to `self` like here...

neurolib/models/model.py
```diff
@@ -2,34 +2,321 @@
...
+        :param append_outputs: append the chunkwise outputs to the outputs attribute, defaults to False, defaults to False
+        :type append_outputs: bool, optional
+        """
+
+        self.initialize_run(bold)
+
+        # override some settings if onedt==True
+        if onedt:
+            self.autochunk(inputs=inputs, append_outputs=append_outputs)
+            return
```
jajcayn: 💥

neurolib/models/model.py
```diff
@@ -2,34 +2,321 @@
...
+            for i, iv in enumerate(self.input_vars):
+                self.params[iv] = inputs[i]
+
+        # run integration
+        self.integrate(append_outputs=append_outputs)
+
+        # reset initial conditions to last state
+        for iv, sv in zip(self.init_vars, self.state_vars):
+            # if state variables are one-dimensional (in space only)
+            if len(self.state[sv].shape) == 1:
```
jajcayn:
```suggestion
            if self.state[sv].ndim == 1:
```

neurolib/utils/functions.py
```diff
@@ -45,6 +45,7 @@ def kuramoto(traces, dt=0.1, smoothing=0.0, peakrange=[0.1, 0.2]):
                 lastMax = m
             phases[n].append(2 * np.pi)
         else:
+            logging.warn("Kuramoto: No peaks found, returning 0.")
```
jajcayn:
```suggestion
            logging.warning("Kuramoto: No peaks found, returning 0.")
```

as per https://stackoverflow.com/a/15655674

tests/test_autochunk.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    def test_onstep_chunkwise(self):
+        """Tests the one-dt (one-step) integration and compares it to normal integration.
+        """
+        ds = Dataset("hcp")
+
+        for model_class in [ALNModel, HopfModel, FHNModel]:
+            logging.info(f"Testing {model_class}...")
+            ModelClass = model_class
```
jajcayn:
```suggestion
```

tests/test_autochunk.py
```diff
@@ -0,0 +1,64 @@
...
+            duration = 200
+            model = ModelClass(Cmat=ds.Cmat, Dmat=ds.Dmat)
```
jajcayn:
```suggestion
            model = model_class(Cmat=ds.Cmat, Dmat=ds.Dmat)
```

neurolib/models/fhn/loadDefaultParams.py
caglorithm: took the default values from `aln` before :) The main reason is because aLN works in `kHz`, hence the very low coupling strength (compared to `Hz` output, or anything really that is of order `1e0` to `1e2`.

neurolib/models/fhn/model.py
caglorithm: nice catch

neurolib/models/aln/model.py
```diff
@@ -46,61 +77,35 @@ def __init__(
...
-                seim,
-                siem,
-                siim,
-                seev,
-                seiv,
-                siev,
-                siiv,
-            ) = ti.timeIntegration(self.params)
+        # Initialize base class Model
+        super().__init__(
```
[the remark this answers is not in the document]
caglorithm: Thanks a lot for this hint, now the init function is very concise :) Fixed in commit 0a860cb4ee3cb28996000d9933491ee6a0fb55a4

neurolib/models/aln/model.py
```diff
@@ -46,61 +77,35 @@ def __init__(
...
+            default_output=self.defaultOutput,
+            bold=bold,
+            name=self.name,
+            description=self.description,
+        )
 
-        # convert output from kHz to Hz
-        rates_exc = rates_exc * 1000.0
-        rates_inh = rates_inh * 1000.0
+    def getMaxDelay(self):
```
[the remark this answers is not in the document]
caglorithm: Docstring already has this hint: 
` Note: Maxmimum delay is given in units of dt.`

neurolib/models/aln/model.py
caglorithm: The parameters `de` and `di` are very specific for the `aln` model, the other models don't have local delays, that's why they can rely on the superclass' `getMaxDelay()`. I didn't want to add another "register your delays" type of functionality, which I why I simply chose to overload the function here. 

Your suggested change is nicer, I will add it! Thanks!

neurolib/models/model.py
caglorithm: That's what the CPU does, yes! :D

neurolib/models/model.py
caglorithm: This tip improved my life by `0.01%`, thank you. Always been doing it wrong.

neurolib/utils/functions.py
caglorithm: Thanks! fixed

neurolib/models/hopf/loadDefaultParams.py
caglorithm: This one should be refactored, I agree. Not gonna do it for this PR, hope we won't forget about it 👍

PR #58, opened 2020-04-03
Thalamic mass model

jajcayn:
Simple thalamic conductance-based mass model, due to Schellenberger Costa et al.
A bit different from other models, since this is a thalamus-only model, i.e. doesn't work within a network. So far without inputs from other areas.

Left to do:
* [ ] example notebook
* [ ] inputs from other (cortical) areas

Reference: https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005022

### DIFF
--- a/neurolib/models/thalamus/timeIntegration.py
+++ b/neurolib/models/thalamus/timeIntegration.py
@@ -57,25 +57,25 @@ def timeIntegration(params):
 
     # model output
     V_t = np.zeros((startind + len(t),))
-    V_t[0] = params["tcr_V_init"]
     V_r = np.zeros((startind + len(t),))
-    V_r[0] = params["trn_V_init"]
     Q_t = np.zeros((startind + len(t),))
     Q_r = np.zeros((startind + len(t),))
     # init
+    V_t[:startind] = params["V_t_init"]
+    V_r[:startind] = params["V_r_init"]
     Ca = params["Ca_init"]
-    h_T_t = 0.0
-    h_T_r = 0.0
-    m_h1 = 0.0
-    m_h2 = 0.0
-    s_et = 0.0
-    s_gt = 0.0
-    s_er = 0.0
-    s_gr = 0.0
-    ds_et = 0.0
-    ds_gt = 0.0
-    ds_er = 0.0
-    ds_gr = 0.0
+    h_T_t = params["h_T_t_init"]
+    h_T_r = params["h_T_r_init"]
+    m_h1 = params["m_h1_init"]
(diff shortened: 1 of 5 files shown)


jajcayn:
hmm `timeIntegration.py` file is not ignored by codecov I think, that is weird...

neurolib/models/thalamus/timeIntegration.py
```diff
@@ -0,0 +1,304 @@
...
+        return g_L * (voltage - E_L)
+
+    def _potassium_leak_current(voltage):
+        return g_LK * (voltage - E_K)
+
+    def _syn_exc_current(voltage, synaptic_rate):
+        return g_AMPA * synaptic_rate * (voltage - E_AMPA)
+
+    def _syn_inh_current(voltage, synaptic_rate):
+        return g_GABA * synaptic_rate * (voltage - E_GABA)
```
caglorithm, line 223: cool 😎

neurolib/models/thalamus/timeIntegration.py
```diff
@@ -0,0 +1,304 @@
...
+        h_T_t = h_T_t + dt * d_h_T_t
+        h_T_r = h_T_r + dt * d_h_T_r
+        m_h1 = m_h1 + dt * d_m_h1
+        m_h2 = m_h2 + dt * d_m_h2
+        s_et = s_et + dt * d_s_et
+        s_gt = s_gt + dt * d_s_gt
+        s_er = s_er + dt * d_s_er
+        s_gr = s_gr + dt * d_s_gr
+        # noisy variable
+        ds_et = ds_et + dt * d_ds_et + gamma_e ** 2 * d_phi * sqrt_dt * noise[i]
```
caglorithm, line 299: `gamma_e_sq = gamma_e ** 2` could be precomputed

jajcayn:
mmm, not sure. the thing is, input from other cortical areas - not sure how to do it :) that's part of the research I guess, so maybe I'd merge it as is so at least the isolated thalamus is there and then we might start to play around with connecting it first to one ALN node (excitatory + inhibitory) and try to reproduce the dynamics as in the source paper of the model (where they used a different model for cortical column) and just then try to connect it to the full network, what do you say?

I just rebased it, waiting for the tests to finish :)

caglorithm:
> not sure how to do it :) that's part of the research I guess

I agree, we can merge this model as is and keep on deving the `Circuit` as we go! Are you going to add a notebook explaining the model? Otherwise we merge after checks are green :)

PR #63, opened 2020-04-08
Exploration tools

caglorithm:
Improvement of the `BoxSearch` Class.
- New Notebook `example-1.2-brain-network-exploration.ipynb`
- New file: `neurolib/optimize/exploration/explorationUtils.py` aka `eu`
- New file: `testing/test_explorationUtils.py`
- `search.run()` passes arguments to `model.run()`.
- `model.params['bold'] = True` forces BOLD simulation
- Postprocess exploration results using `eu.processExplorationResults()`
- Find simulation parameters from `search.dfResults` using `eu.findCloseResults()`
- Plot exploration results via `eu.plotExplorationResults()` (example attached below)
- `Model` BOLD output time array renamed to `t_BOLD`
- Models can now generate random initial conditions using `model.randomICs()`
- Version bump to `v0.4.2`

image


neurolib/models/model.py
```diff
@@ -179,7 +177,23 @@ def run(
...
                 bold = False
             self.integrateChunkwise(chunksize=chunksize, bold=bold, append_outputs=append)
-            return
+
+        # check if there was a problem with the simulated data
+        self.checkOutputs()
+
+    def checkOutputs(self):
+        # check nans in output
+        if np.isnan(np.sum(self.output)):
```
jajcayn, line 186:
```suggestion
        if np.isnan(self.output).any():
```
maybe cleaner? not sure...

neurolib/models/model.py
```diff
@@ -179,7 +177,23 @@ def run(
...
+            logging.error("nan in model output!")
+        else:
+            EXPLOSION_THRESHOLD = 1e20
+            if np.sum(self.output > EXPLOSION_THRESHOLD) > 0:
```
jajcayn, line 190:
```suggestion
            if (self.output > EXPLOSION_THRESHOLD).any():
```
maybe cleaner?

neurolib/models/model.py
```diff
@@ -179,7 +177,23 @@ def run(
...
+                logging.error("nan in model output!")
+
+        # check nans in BOLD
+        if "BOLD" in self.outputs:
+            if np.isnan(np.sum(self.outputs.BOLD.BOLD)):
```
jajcayn, line 195:
```suggestion
            if np.isnan(self.outputs.BOLD.BOLD).any():
```

neurolib/models/model.py
```diff
@@ -179,7 +177,23 @@ def run(
...
+                logging.error("nan in BOLD output!")
```
jajcayn, line 196: the logging errors just write an error, is that ok? don't you want to throw an actual error or exception?

neurolib/optimize/exploration/exploration.py
```diff
@@ -54,38 +61,46 @@ def __init__(self, model=None, parameterSpace=None, evalFunction=None):
         # todo: use random ICs for every explored point or rather reuse the ones that are generated at model initialization
         self.useRandomICs = False
 
+        if filename is None:
+            filename = "exploration.hdf"
```
jajcayn, line 65:
```suggestion
        filename = filename or "exploration.hdf"
```

tests/test_explorationUtils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def test_processExplorationResults(self):
+        self.search.dfResults = eu.processExplorationResults(
+            self.search.results, self.search.dfResults, model=self.model, ds=self.ds, bold_transient=0
+        )
+
+    def test_findCloseResults(self):
+        eu.findCloseResults(self.search.dfResults, dist=1, x_ext=0, K_gl=0.0)
+
+    @pytest.mark.skipif(sys.platform == "darwin", reason="plotting does not work on macOS")
```
jajcayn, line 65: nice! didn't know this is possible ;)

neurolib/models/model.py
caglorithm: Yes, thanks for making me aware of this. This is the way!

neurolib/optimize/exploration/exploration.py
caglorithm: This is very ninja, never saw it! Thanks!

neurolib/models/model.py
caglorithm: Yep, I'm aware of this. I don't want to raise an exception though, because the original signal (not bold) could still be fine and I don't want to upset `pypet` during explorations... :/

jajcayn: that's true, good catch... maybe I would then set full BOLD time series to NaNs... just in case

caglorithm: Sometimes, the bold signal "crashes" mid simulation when for example the system undergoes a random bifurcation (jump from down to up). Often, _some_ notes will go `nan` and others still work because of their connection strengths. To be able to debug these cases, I'll let it just throw this message for now.... 

I wanna implement a second BOLD mechanism using the lowpass-filtered envelope of the signal some day, hope that it won't have these issues anymore.

PR #64, opened 2020-04-08
thalamus: firing rates kHz -> Hz conversion

jajcayn:
quickie :)


neurolib/models/thalamus/timeIntegration.py
```diff
@@ -301,4 +301,8 @@ def _syn_inh_current(voltage, synaptic_rate):
         ds_er = ds_er + dt * d_ds_er
         ds_gr = ds_gr + dt * d_ds_gr
 
+    # kHz -> Hz for firing rates
+    Q_t *= 1000.0
```
caglorithm, line 305:
```suggestion
    Q_t = np.sum([Q_t] * 1000.0)
```
How about this hack I found on stackoverflow 💩

PR #70, opened 2020-04-12
Extended Wilson-Cowan model 

ChristophMetzner:
Extended three-population Wilson-Cowan model. The model consists of one excitatory and two inhibitory population, one representing dendrite-targeting SST+ cells and the other soma-targeting PV+ cells. Furthermore, the PV+ inhibitory population can produce subtractive or divisive inhibition, or a mixture thereof. 

Taken from: Papasavvas et al., *Divisive gain modulation enables flexible and rapid entrainment in a
neocortical microcircuit model*, J. Neurophysiol., 2020, https://doi.org/10.1152/jn.00401.2019


neurolib/models/subdivwc/timeIntegration.py
```diff
@@ -0,0 +1,293 @@
...
+            inh_d_input_d[no] = 0
+
+            for l in range(N):
+                exc_input_d[no] += K_gl * Cmat[no, l] * (excs[l, i - Dmat_ndt[no, l] - 1])
+
+
+
+            # subtractive/divisive Wilson-Cowan model
+            exc_rhs = (
+                1/tau_exc*(-excs[no, i - 1]+(k_exc-excs[no, i - 1])*F_exc(w1*excs[no, i - 1]+exc_ext[no],
```
jajcayn, line 270:
```suggestion
                1/tau_exc*(-excs[no, i - 1]+(k_exc-excs[no, i - 1])*F_exc(w1*excs[no, i - 1]+exc_ext[no] + exc_input_d ,
```
maybe `+ exc_input_d`? otherwise I do not see any use of network input `exc_input_d ` :)

jajcayn: by the way I went and check basic 2population WC model already in `neurolib` and I think I found a bug.
In the original WC model (`neurolib/models/wc/timeIntegration.py`) both excitatory and inhibitory population got ` + exc_input_d` into sigmoid transfer function and I feel from the WC definition, only excitatory should get network input from other excitatory nodes

caglorithm: > by the way I went and check basic 2population WC model already in `neurolib` and I think I found a bug.
> In the original WC model (`neurolib/models/wc/timeIntegration.py`) both excitatory and inhibitory population got ` + exc_input_d` into sigmoid transfer function and I feel from the WC definition, only excitatory should get network input from other excitatory nodes

Thank for finding the bug! We need to fix it!

jajcayn: not sure if it's a bug, I worked with WC model few years back, so it's better if @ChristophMetzner confirms :)

caglorithm:
Thanks a lot for the PR! Interesting model! Are you working with this model already? 

We should start thinking about documenting each model a bit better since we're getting more and more :) Maybe this is a good opportunity to start describing models in the `model.py` header comment and also include the references where the model is from? What do you think @jajcayn?

Also, it doesn't make sense to replicate the minimal Notebook for every new model but we could focus on what the special features of a model interesting etc. An overview of this model for example should provide insights in its dynamical properties, especially due to the second inhibitory population. I would appreciate a better documentation before merging this PR!

neurolib/models/subdivwc/timeIntegration.py
caglorithm: Yes, we should wait for @ChristophMetzner. However, every implementation of a brain network that I've seen so far only uses excitatory coupling.

jajcayn:
I agree! My take would be something like:
```
class ALNModel(Model):
    """
    Mean-field approximation of adaptive exponential integrate-and-fire neuron. The Fokker-Planck mean-field equation is solved using a precomputed linear-nonlinear cascade model. This model is usually used for cortical nodes within a whole-brain network. Each node is modelled as two populations of ALN neural masses, one excitatory with adaptive current, one inhibitory without adaptive current. The brain network is realised by coupling excitatory to excitatory masses within the network using the structural connectivity matrix and supports network delays.

    References:
        * first paper
        * second paper if applicable
        * etc.
    """
```

so I'd write a few sentences on what is the model actually modelling, is it suitable for whole-brain net, how is the network coupling actually realised and some references (e.g. on for general equations, one for widely accepted parameters, etc)...

as for the example notebooks, I'd do one notebook per model, i.e. WC models (2 of them) would have one notebook - basic usage of 1 node of WC, a network of WC and then differences between 2pop WC and extended 3pop WC model.

neurolib/models/subdivwc/timeIntegration.py
ChristophMetzner: Thanks for finding that bug/those bugs. You're right, it should be `+ exc_input_d`.
Wrt to the original WC model, I think that you're right that most (all?) people only give input from other network nodes to the excitatory population and we should stick to that.

ChristophMetzner: I will fix those bugs asap (probably tomorrow).

ChristophMetzner:
Good points. I like the idea of one notebook per model/model type. I have searched the literature for different types of WC models and will probably implement 2-3 more over the next few weeks, all different extensions to the original model (e.g. synaptic rise/decay times, two inhibitory populations, subtractive/divisive inhibition, neuromodulation). Then, one large notebook presenting all of them and highlighting the differences and commonalities makes a lot of sense.

I also agree with the more detailed explanations in the model header. Will update the two models accordingly.

PR #71, opened 2020-05-11
Exploration and Evolution: improvements

caglorithm:
**Evolution:** 
* NSGA-2 algorithm implemented (Deb et al. 2002)
* Preselect algorithms (using `algorithm="adaptive"` or `"nsga2"`)
* Custom operators for all evolutionary operations possible
* Keep track of the evolution history using `evolution.history`
* `networkx` compatible genealogy `evolution.tree` available from `evolution.buildEvolutionTree()` [1]
* `saveEvolution()` and `loadEvolution()` using `dill` [2]
* `evolution.dfPop` now has all fitness values as well
* get scores using `getScores()`
* `evolutionaryUtils.plotProgress()` added
* `evolutionaryUtils. plotPopulation()` better scatterplots with color option `evolution.plotColor`

**Exploration:**
* `loadResults()` has memory cap to avoid filling up RAM
* `loadDfResults()` creates the parameter table from a previous simulation
* New method `aggregateResultsToDfResults()` stores simulation results (into `dfResults`) by either loading results from disk or from memory, if `loadResults()` was called before.
* ℹ The user does not need to know the difference. In short, either use `loadResults(all=True)` to load everything into memory (available as `.results`) or use `all=False` to load runs individually from hdf. Both options populate `dfResults`.
* `explorationUtils.plotExplorationResults()` for plotting exactly that, with some advanced functions like alpha maps and contours for regions

**devUtils**
* Initially forgot about this file but hell, why not, it's a hacker file with some neat features
   * `plot_outputs()` like a true numerical simlord
   * `model_fit()` to compute the model's FC and FCD fit to the dataset, could be usefull for everyone
   * `getPowerSpectrum()` sais everything
   * `getMeanPowerSpectrum()` yes
   *  a very neat `rolling_window()` from a `numpy` PR that never got accepted

**Other:**
* Data loading:
    * `Dataset` can load different SC matrix normalizations: `"max", "waytotal", "nvoxel"`
    * Can precompute FCD matrices to avoid having to do it later (`fcd=True`)
* `neurolib/utils/atlas.py` added with aal2 region names (thanks @jajcayn) and coordinates of centers of regions (from scans of my own brain 🤯)
* `ParameterSpace` has `.lowerBound` and `.upperBound`.
* `pypet` finally doesn't create a billion log files anymore due to a custom log config

[1] Plotting the genealogy tree:
```python
import matplotlib.pyplot as plt
import networkx as nx
from networkx.drawing.nx_pydot import graphviz_layout

G = nx.DiGraph(evolution.tree)
G = G.reverse()     # Make the graph top-down
pos = graphviz_layout(G, prog='dot')
plt.figure(figsize=(8, 8))
nx.draw(G, pos, node_size=50, alpha=0.5, node_color=list(evolution.id_score.values()), with_labels=False)
plt.show()
```
tree

[2] Loading precomputed evolution
```python
evaluateSimulation = lambda x: x # the funciton can be ommited, that's why we define a lambda here
pars = ParameterSpace(['a', 'b'], # should be same as previously saved evolution
                      [[0.0, 4.0], [0.0, 5.0]])
evolution = Evolution(evaluateSimulation, pars, weightList = [1.0])
evolution = evolution.loadEvolution("data/evolution-results-2020-05-15-00H-24M-48S.dill")
```

Todo: 
* [x] description of PR
* [ ] improve coverage


neurolib/optimize/evolution/evolution.py
```diff
@@ -360,14 +399,16 @@ def evalPopulationUsingPypet(self, traj, toolbox, pop, gIdx):
             # store outputs of simulations in population
             pop[idx].fitness.values = fitnessesResult
             # mean fitness value
-            pop[idx].fitness.score = np.nansum(pop[idx].fitness.wvalues) / (len(pop[idx].fitness.wvalues))
+            #pop[idx].fitness.score = np.nansum(pop[idx].fitness.wvalues) / (len(pop[idx].fitness.wvalues))
+            # works with nans and infs:
+            pop[idx].fitness.score = np.ma.masked_invalid(pop[idx].fitness.wvalues).sum() / (len(pop[idx].fitness.wvalues))
```
caglorithm, line 405: This way the sum works with `np.nan` and `np.inf`.

neurolib/optimize/evolution/evolution.py
```diff
@@ -360,14 +399,16 @@ def evalPopulationUsingPypet(self, traj, toolbox, pop, gIdx):
...
         return pop
 
     def getValidPopulation(self, pop):
-        return [p for p in pop if not np.any(np.isnan(p.fitness.values))]
+        return [p for p in pop if not (np.isnan(p.fitness.values).any() or np.isinf(p.fitness.values).any()) ]
```
caglorithm, line 409: `np.inf` is now also an invalid result (sometimes preferable to `np.nan` in case you have Multi Objective Optimization, you might have one invalid result as `+np.inf` and the other as `-np.inf`, this doesn't work with `np.nan`)

neurolib/optimize/evolution/evolution.py
```diff
@@ -414,8 +455,9 @@ def runInitial(self):
         # save all simulation data to pypet
         self.pop = eu.saveToPypet(self.traj, self.pop, self.gIdx)
 
-        # Only the best indviduals are selected for the population the others do not survive
-        self.pop[:] = self.toolbox.selBest(self.pop, k=self.traj.popsize)
+        # reduce initial population to popsize
+        #self.pop = self.toolbox.select(self.pop, k=self.traj.popsize, **self.SELECT_P)
+
```
caglorithm, line 461: We do not need to reduce the initial population here because it will be done in the first evolution round anyway....

neurolib/optimize/evolution/evolution.py
```diff
@@ -434,21 +476,19 @@ def runEvolution(self):
...
 
             # ------- Create the next generation by crossover and mutation -------- #
             ### Select parents using rank selection and clone them ###
-            offspring = list(map(self.toolbox.clone, self.toolbox.selRank(self.pop, self.POP_SIZE, self.RANKP)))
-            # n_offspring = min(len(validpop), self.POP_SIZE)
-            # if n_offspring % 2 != 0:
-            #     n_offspring -= 1
-            # offspring = list(map(self.toolbox.clone, self.toolbox.selRank(validpop, n_offspring)))
+            offspring = list(map(self.toolbox.clone, self.toolbox.selectParents(self.pop, self.POP_SIZE, **self.PARENT_SELECT_P)))
+
```
caglorithm, line 488: kwargs for evolution operators are handled this way

neurolib/optimize/evolution/evolution.py
```diff
@@ -484,19 +524,25 @@ def runEvolution(self):
...
+                self.info(plot=True, info=False)
+                # # population summary
+                # eu.printParamDist(self.pop, self.paramInterval, self.gIdx)
+                # bestN = 5
+                # print(f"Best {bestN} individuals:")
+                # eu.printIndividuals(self.toolbox.selBest(self.pop, bestN), self.paramInterval)                
+                # # plotting
+                # eu.plotPopulation(
+                #     self.pop, self.paramInterval, self.gIdx, plotScattermatrix=True, save_plots=self.trajectoryName
+                # )
```
caglorithm, line 546: I should get rid of these comments...

neurolib/optimize/exploration/exploration.py
```diff
@@ -249,44 +269,83 @@ def run(self, **kwargs):
         assert self.initialized, "Pypet environment not initialized yet."
         self.env.run(self.evalFunction)
 
-    def loadResults(self, filename=None, trajectoryName=None, pypetShortNames=True):
+    def loadResults(self, all=True, filename=None, trajectoryName=None, pypetShortNames=True, memory_cap=95.0):
```
caglorithm, line 272: `memory_cap` avoids filling up RAM while loading results.

neurolib/optimize/exploration/exploration.py
```diff
@@ -249,44 +269,83 @@ def run(self, **kwargs):
...
+                    for key, value in r.items():
+                        new_key = "".join(key.split(".", 2)[2:])
+                        new_dict[new_key] = r[key]
+                    self.results[i] = copy.deepcopy(new_dict)
+
+        self.aggregateResultsToDfResults()
 
         logging.info("All results loaded.")
 
+    def aggregateResultsToDfResults(self):
```
caglorithm, line 325: New function to aggregate all simulation results to the pandas dataframe. Can be used without loading all results to RAM but loading (and unloading) them one by one using `getRun()`

neurolib/optimize/exploration/exploration.py
```diff
@@ -333,3 +392,32 @@ def getRun(self, runId, filename=None, trajectoryName=None, pypetShortNames=True
             pypetTrajectory = pu.loadPypetTrajectory(filename, trajectoryName)
 
         return pu.getRun(runId, pypetTrajectory, pypetShortNames=pypetShortNames)
+
+    def getResult(self, runId):
```
caglorithm, line 396: Good guy

neurolib/utils/devutils.py
```diff
@@ -1,40 +1,391 @@
 import matplotlib.pyplot as plt
 import numpy as np
+import scipy
 
 import neurolib.utils.functions as func
 
 
-def plot_outputs(model, bold_transient=10000):
-    _, axs = plt.subplots(2, 3, figsize=(12, 6))
+def plot_outputs(
```
caglorithm, line 8: I use this file only for my personal plotting purposes etc....

neurolib/utils/loadData.py
```diff
@@ -15,7 +15,7 @@ class Dataset:
     """Dataset class.
     """
 
-    def __init__(self, datasetName=None, normalizeCmats="max"):
+    def __init__(self, datasetName=None, normalizeCmats="max", fcd=False):
```
caglorithm, line 18: Can now precompute FCD matrices or not. Added the option because it can take quite some time.

neurolib/utils/loadData.py
```diff
@@ -179,7 +218,7 @@ def _loadSubjectFiles(self, dsBaseDirectory, filterSubcorticalAAL2=False):
                     continue
                 for f in fnames:
                     # dirty
-                    subject = f.split("/")[-3]
+                    subject = f.split(os.path.sep)[-3]
```
caglorithm, line 221: For Windows compatibility xD (found by Lena, our only Windows user :D)

neurolib/utils/paths.py
```diff
@@ -4,3 +4,5 @@
 
 HDF_DIR = "./data/hdf/"
 FIGURES_DIR = "./data/figures/"
+
+PYPET_LOGGING_CONFIG = "./neurolib/utils/pypet_logging.ini"
```
caglorithm, line 8: Pypet finally doesn't log in files anymore using this custom log.ini

requirements.txt
```diff
@@ -1,3 +1,4 @@
+psutil
```
caglorithm, line 1: For getting RAM usage in `exploration.loadResults()`

neurolib/optimize/evolution/evolution.py
jajcayn: really not needed, but for some reason I do not like `pop[idx].something.something` repeated over and over... I'd do simple class method something like
```python
def _set_population(self, pop, output, fitness, score):
    pop.outputs = output
    pop.fitness.value = fitness
    pop.fitness.score = score

# and actually call it as
self._set_population(pop[idx], returnedOutputs, returnedOutputs, np.ma.masked_invalid(pop[idx].fitness.wvalues).sum() / (len(pop[idx].fitness.wvalues)))
```

neurolib/optimize/evolution/evolution.py
jajcayn: yep :)

neurolib/optimize/evolution/evolution.py
```diff
@@ -577,6 +679,12 @@ def loadResults(self, filename=None, trajectoryName=None):
             filename = self.HDF_FILE
         self.traj = pu.loadPypetTrajectory(filename, trajectoryName)
 
+    def getScores(self):
+        """Returns the scores of the current valid population
+        """
+        validPop = self.getValidPopulation(self.pop)
+        return np.array([validPop[i].fitness.score for i in range(len(validPop))])
```
jajcayn, line 687:
```suggestion
        return np.array([pop.fitness.score for pop in validPop])
```

neurolib/optimize/evolution/evolutionaryUtils.py
```diff
@@ -134,18 +122,17 @@ def plotPopulation(
 
     # Gather all the fitnesses in one list and print the stats
     # validPop = [p for p in pop if not np.isnan(p.fitness.score)]
-    validPop = [p for p in pop if not np.any(np.isnan(p.fitness.values))]
+    # validPop = [p for p in pop if not np.any(np.isnan(p.fitness.values))]
+    validPop = [p for p in pop if not (np.isnan(p.fitness.values).any() or np.isinf(p.fitness.values).any()) ]
```
jajcayn, line 140: I think I've seen this checking whether a population is valid, what about a short convenience function?
```python
def _is_pop_valid(pop):
    return not (np.isnan(pop.fitness.values).any() or np.isinf(pop.fitness.values).any())
```
then just
```python
validPop = [p for p in pop if _is_pop_valid(p)]
```
and also in other places where you check validness of the population

neurolib/optimize/evolution/evolutionaryUtils.py
```diff
@@ -134,18 +122,17 @@ def plotPopulation(
...
     popArray = np.array([p[0 : len(paramInterval._fields)] for p in validPop]).T
     scores = np.array([validPop[i].fitness.score for i in range(len(validPop))])
```
jajcayn, line 142:
```suggestion
    scores = np.array([pop.fitness.score for pop in validPop])
```

neurolib/optimize/evolution/evolutionaryUtils.py
```diff
@@ -191,8 +184,12 @@ def printEvolutionInfo(evolution):
     print(f"Number of generations: {evolution.NGEN}")
```
jajcayn, line 198: maybe all `print` -> `logging.info` ?

neurolib/optimize/exploration/exploration.py
```diff
@@ -333,3 +402,32 @@ def getRun(self, runId, filename=None, trajectoryName=None, pypetShortNames=True
...
+            result = self.results[runId]        
+        else:
+            # or from disk if results haven't been loaded yet
+            result = self.getRun(runId)
+        return result
+
+    def info(self):
+        """Print info about the current search.
+        """
+        print(
```
jajcayn, line 425: I think this is exactly what predefined `__str__(self)` functions are for in classes:)

neurolib/optimize/exploration/explorationUtils.py
```diff
@@ -36,79 +46,119 @@ def plotExplorationResults(
...
+        # pivot data and plot
+        df_pivot = df.pivot_table(values=plot_key, index=par2, columns=par1, dropna=False)
 
-        if one_figure == False:
-            fig = plt.figure(figsize=(5, 4), dpi=150)
-            ax = plt.gca()
+        plot_clim = (np.nanmin(df_pivot.values), np.nanmax(df_pivot.values))
+
+        if symmetric_colorbar:
+            plot_clim = (-np.nanmax(df_pivot.values), np.nanmax(df_pivot.values))
```
jajcayn, line 92: what if abs(min_value) > max_value? I think better is

```suggestion
            plot_clim = (-np.max(np.abs(plot_clim)), np.max(np.abs(plot_clim)))
```

neurolib/optimize/exploration/explorationUtils.py
```diff
@@ -119,23 +169,145 @@ def plotExplorationResults(
...
+    cmap = plt.cm.plasma
+    colors = Normalize(np.nanmin(image), np.nanmax(image), clip=True)(image)
+    colors = cmap(colors)
+    colors[..., -1] = alphas
+    return colors
+
+def plotResult(search, runId, z_bold = False, **kwargs):
+    fig, axs = plt.subplots(1, 3, figsize=(8, 2), dpi=300, gridspec_kw={'width_ratios': [1, 1.2, 2]})
+
+    bold_transient = int(kwargs["bold_transient"] / 2) if "bold_transient" in kwargs else int(30 / 2)
```
jajcayn, line 203:
```suggestion
    bold_transient = int(kwargs["bold_transient"] / 2) if "bold_transient" in kwargs else 30 // 2
```

jajcayn: int division is `//`

neurolib/utils/atlases.py
```diff
@@ -0,0 +1,425 @@
...
+         78.7233908 ,  76.49070506,  67.14054291,  63.17929148,
+         41.71139163,  32.11614479,  40.15115331,  34.04905621,
+         42.64541727,  41.04761205,  53.79855827,  51.72007042,
+         76.8045977 ,  86.64375174,  87.16152042,  70.76570352,
+         56.91432792,  49.77622217,  52.03441237,  59.375     ]])
+
+    def __init__(self):
+        super().__init__(self.aal2, coords=self.aal2_centers)
+        # define convenience regions - indexing from 0!!
+        #self.cortex = [region for region in range(94) if region not in np.array(list(range(40, 46)) + list(range(74, 82)))]
```
jajcayn, line 323: remove commented code?

neurolib/utils/devutils.py
jajcayn: wunderbar! I won't review it then :)

neurolib/utils/loadData.py
jajcayn: 🤯 Windows🤯

neurolib/optimize/evolution/evolutionaryUtils.py
caglorithm: This one is truly for printing, will be called only for `verbose=True`, otherwise it could be suppressed if the logger is configured differently...

PR #84, opened 2020-06-04
Load evolution from disk

caglorithm:
* Load an evolution from disk that was previously saved using `dill`.
Saving:
```python
... # evolution done here
evolution.saveEvolution("data/evolution.dill")
```
Loading, in another session:
```python
# initialize object
evolution = Evolution(lambda x: x, ParameterSpace(['mock'],  [[0, 1]]))
# load from disk
evolution = evolution.loadEvolution("data/evolution.dill")
```

Because the objects are not serializable, I needed to use `dill` for saving and loading. This means, this would be a new dependency (which I' obviously trying to avoid as much as possible). At the same time, saving and loading a previous evolution / exploration is, I think, a very important feature and reinitializing the evolution with the loaded dill is a bit non-trivial, so I thought it makes sense to solve this issue for the user instead of letting them figure it out themselves. I'd be happy about comments on this (@jajcayn?).

* Pandas DataFrame with all individuals from the entire evolution
```python
evolution.dfEvolution
```
Sorry:
Most of the changes are actually black formatting.... :( more coming very soon.

Todo:
* [x] Testing of `saveEvolution()` and `loadEvolution()`
* [ ] Lazy import of dill, is it ok? Should give `ImportError` if import fails
* [x] Docstring dfEvolution

### DIFF
--- a/neurolib/optimize/evolution/evolution.py
+++ b/neurolib/optimize/evolution/evolution.py
@@ -36,7 +36,7 @@ class Evolution:
         POP_INIT_SIZE=100,
         POP_SIZE=20,
         NGEN=10,
-        algorithm='adaptive',
+        algorithm="adaptive",
         matingOperator=None,
         MATE_P=None,
         mutationOperator=None,
@@ -46,7 +46,7 @@ class Evolution:
         parentSelectionOperator=None,
         PARENT_SELECT_P=None,
         individualGenerator=None,
-        IND_GENERATOR_P=None
+        IND_GENERATOR_P=None,
     ):
         """Initialize evolutionary optimization.
         :param evalFunction: Evaluation function of a run that provides a fitness vector and simulation outputs
@@ -93,10 +93,14 @@ class Evolution:
         """
 
         if weightList is None:
-            logging.info("weightList not set, assuming single fitness value to be maximized.")
+            logging.info(
+                "weightList not set, assuming single fitness value to be maximized."
+            )
             weightList = [1.0]
 
-        trajectoryName = "results" + datetime.datetime.now().strftime("-%Y-%m-%d-%HH-%MM-%SS")
+        trajectoryName = "results" + datetime.datetime.now().strftime(
+            "-%Y-%m-%d-%HH-%MM-%SS"
+        )
         logging.info(f"Trajectory Name: {trajectoryName}")
         self.HDF_FILE = os.path.join(paths.HDF_DIR, filename)
         trajectoryFileName = self.HDF_FILE
@@ -116,7 +120,7 @@ class Evolution:
             multiproc=True,
             ncores=ncores,
             complevel=9,
-            log_config=paths.PYPET_LOGGING_CONFIG,            
+            log_config=paths.PYPET_LOGGING_CONFIG,
         )
 
         # Get the trajectory from the environment
@@ -130,7 +134,6 @@ class Evolution:
         self.model = model
         self.evalFunction = evalFunction
         self.weightList = weightList
-
 
         self.NGEN = NGEN
         assert POP_SIZE % 2 == 0, "Please chose an even number for POP_SIZE!"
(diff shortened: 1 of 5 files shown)


neurolib/optimize/evolution/evolution.py
```diff
@@ -700,34 +802,113 @@ def loadEvolution(self, fname):
...
+        # add fitness columns
+        # NOTE: have to do this with wvalues and divide by weights later, why?
+        # Because after loading the evolution with dill, somehow multiple fitnesses
+        # dissappear and only the first one is left. However, wvalues still has all
+        # fitnesses, and we have acces to weightList, so this hack kind of helps
+        n_fitnesses = len(self.pop[0].fitness.wvalues)
+        for i in range(n_fitnesses):
+            for ip, p in enumerate(allIndividuals):
+                column_name = "f" + str(i)
+                dfEvolution.loc[ip, column_name] = (
```
jajcayn, line 906:
```suggestion
                dfEvolution.loc[ip,f"f{i}"] = (
```

neurolib/optimize/evolution/evolution.py
```diff
@@ -700,34 +802,113 @@ def loadEvolution(self, fname):
...
+
+        # add fitness columns
+        # NOTE: have to do this with wvalues and divide by weights later, why?
+        # Because after loading the evolution with dill, somehow multiple fitnesses
+        # dissappear and only the first one is left. However, wvalues still has all
+        # fitnesses, and we have acces to weightList, so this hack kind of helps
+        n_fitnesses = len(self.pop[0].fitness.wvalues)
+        for i in range(n_fitnesses):
+            for ip, p in enumerate(allIndividuals):
+                column_name = "f" + str(i)
```
jajcayn, line 905:
```suggestion
```

neurolib/optimize/evolution/evolutionaryUtils.py
```diff
@@ -11,18 +11,32 @@
 def saveToPypet(traj, pop, gIdx):
     try:
         traj.f_add_result_group("{}.gen_{:06d}".format("evolution", gIdx))
-        traj.f_add_result("{}.gen_{:06d}.fitness".format("evolution", gIdx), np.array([p.fitness.values for p in pop]))
-        traj.f_add_result("{}.gen_{:06d}.scores".format("evolution", gIdx), np.array([p.fitness.score for p in pop]))
-        traj.f_add_result("{}.gen_{:06d}.population".format("evolution", gIdx), np.array([list(p) for p in pop]))
-        traj.f_add_result("{}.gen_{:06d}.ind_ids".format("evolution", gIdx), np.array([p.id for p in pop]))
+        traj.f_add_result(
+            "{}.gen_{:06d}.fitness".format("evolution", gIdx),
```
jajcayn, line 15: I'd use either `str.format` or f-string in the whole repo, not mixing the two if not necessary...

neurolib/optimize/evolution/evolution.py
```diff
@@ -700,34 +802,113 @@ def loadEvolution(self, fname):
         :rtype: self
         """
         import dill
+
         evolution = dill.load(open(fname, "rb"))
-        evolution.__init__(lambda x: x, self.parameterSpace)
+
+        # parameter space is not saved correctly in dill, don't know why
```
jajcayn, line 808: this is weird...do you have any idea why? what's the problem when you load with `dill`?

neurolib/optimize/evolution/evolution.py
```diff
@@ -678,8 +777,11 @@ def saveEvolution(self, fname=None):
         :type fname: str, optional
         """
         import dill
-        fname = fname or os.path.join("data/", "evolution-" + self.trajectoryName + ".dill")
-        dill.dump(self, open(fname, "wb"))        
+
+        fname = fname or os.path.join(
+            "data/", "evolution-" + self.trajectoryName + ".dill"
+        )
+        dill.dump(self, open(fname, "wb"))
```
jajcayn, line 784: are you trying to dump whole class? maybe try `self.__dict__`... each class have `__dict__` method which is a dictionary of all attributes... when I am saving whole class from delfi, I save it to pickle (that shouldn't matter whether `pickle` or `dill`) with `dump(self.__dict__, fname)`

caglorithm: pickle works only in serializable data.... I forgot which but one of the many objects (pypet?) didn't quite like to be pickled unfortunately.

neurolib/optimize/evolution/evolutionaryUtils.py
caglorithm: completely! this is a artifact from my pre-fstring life, oops

neurolib/optimize/evolution/evolution.py
jajcayn: yes, you're right, might be something is not pickable, then dill is indeed an option ;)
still, I'd try dilling `self.__dict__`, maybe that'll work

caglorithm: Referring to `self.__dict__`, how would you load the original instance back into memory and replace it with `self`?

caglorithm: Guess I'm gonna leave it as is for now, since it seems to work!

PR #88, opened 2020-06-11
evolution: handling of seed = None for pypet

caglorithm:
`pypet` doesn't like to add `None` results. The RNG seed parameter `model.params["seed"]` is usually `None` and was set to `0` to avoid issues before. Now it stays `None` but when saving to `pypet`, the case is handled.

The rest of the other changes are only black formatting...

### DIFF
--- a/neurolib/models/wc/loadDefaultParams.py
+++ b/neurolib/models/wc/loadDefaultParams.py
@@ -23,12 +23,7 @@ def loadDefaultParams(Cmat=None, Dmat=None, seed=None):
     params.dt = 0.1  # ms 0.1ms is reasonable
     params.duration = 2000  # Simulation duration (ms)
     np.random.seed(seed)  # seed for RNG of noise and ICs
-    # set seed to 0 if None, pypet will complain otherwise
-    params.seed = seed or 0
-
-    # make sure that seed=0 remains None
-    if seed == 0:
-        seed = None
+    params.seed = seed
 
     # ------------------------------------------------------------------------
     # global whole-brain network parameters
--- a/neurolib/optimize/evolution/evolution.py
+++ b/neurolib/optimize/evolution/evolution.py
@@ -897,6 +897,11 @@ class Evolution:
         dfEvolution["id"] = indIds
         dfEvolution["gen"] = [p.gIdx for p in allIndividuals]
 
+        # the history keeps all individuals of all generations
+        # there can be duplicates (in elitism for example), which we filter
+        # out for the dataframe
+        dfEvolution = dfEvolution.drop_duplicates()
+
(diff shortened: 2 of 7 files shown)


neurolib/optimize/evolution/evolution.py
```diff
@@ -315,22 +341,33 @@ def initPypetTrajectory(self, traj, paramInterval, POP_SIZE, NGEN, model):
...
         # if a model was given, save its parameters
         # NOTE: Convert model.params to dict() since it is a dotdict() and pypet doesn't like that
         if model is not None:
-            traj.f_add_result("params", dict(model.params), comment="Default parameters")
+            params_dict = dict(model.params)
+            # replace all None with zeros, pypet doesn't like None
+            for key, value in params_dict.items():
+                if value is None:
+                    params_dict[key] = 0
+            traj.f_add_result("params", params_dict, comment="Default parameters")
```
caglorithm, line 362: This is the part that is new

neurolib/models/wc/loadDefaultParams.py
```diff
@@ -23,12 +23,7 @@ def loadDefaultParams(Cmat=None, Dmat=None, seed=None):
     params.dt = 0.1  # ms 0.1ms is reasonable
     params.duration = 2000  # Simulation duration (ms)
     np.random.seed(seed)  # seed for RNG of noise and ICs
-    # set seed to 0 if None, pypet will complain otherwise
-    params.seed = seed or 0
-
-    # make sure that seed=0 remains None
-    if seed == 0:
-        seed = None
```
caglorithm, line 31: and this is what I got rid of in every `loadDefaultParams.py`

neurolib/optimize/evolution/evolution.py
jajcayn: so None seed is saved to pypet as 0 seed. Am I right? What if you actually use 0 seed, that one is also saved as 0 to pypet?

caglorithm: That's still an unresolved issue tbh... Was an issue before but I didn't want to spend too much time with it.

jajcayn: what if seed specifically is saves as string? `str(<any_integer>) = '<any_integer>'` and `str(None) = 'None'`.. than on loading you can simply:
```python
try:
    seed = int(<saved_in_pypet>)
except ValueError:
    seed = None

```

jajcayn: AFAIK `pypet` can do strings...

caglorithm: it does, nice!

PR #95, opened 2020-07-01
MultiModel PR2: base models

jajcayn:
The second PR of the MultiModel framework, now we know how the internals and model building works so this one should be quick -- introducing different models:
* FitzHugh-Nagumo
* thalamic mass model due to Costa et al.
* Wilson-Cowan
* Wong-Wang (both excitatory vs inhibitory version of `Node` - as Wilson-Cowan and Reduced version of `Node` with one mass - as FitzHugh-Nagumo)

(Side note - our holy grail - AdEx model - will follow in PR3 as it is more complicated to define since it involves table lookup and it's tricky to do that in symbolic derivatives).

All models are tested - whether they can be initialised, they are run with both backend and their output compared and moreover they are tested against `neurolib`'s native implementation (with exception of Wong-Wang since it's not in the `neurolib`)

WIP:
* [x] Wong-Wang testing against virtual brain implementation - they have reduced version
* [x] finish unit tests for Wong-Wang

I suggest you can start reviewing other models than Wong-Wang and in the meantime, I'll compare this implementation with TVB's

### DIFF
--- a/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
+++ b/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
@@ -128,8 +128,7 @@ class FitzHughNagumoNetwork(Network):
 
     sync_variables = ["network_x", "network_y"]
     # define default coupling in FitzHugh-Nagumo network
-    x_coupling = "diffusive"
-    y_coupling = "none"
+    default_coupling = {"network_x": "diffusive", "network_y": "none"}
 
     def __init__(
         self, connectivity_matrix, delay_matrix, mass_params=None, seed=None,
@@ -169,6 +168,3 @@ class FitzHughNagumoNetwork(Network):
         assert all(all_couplings[0] == coupling for coupling in all_couplings)
         # invert as to name: idx
         self.coupling_symbols = {v: k for k, v in all_couplings[0].items()}
-
-    def _sync(self):
-        return self._couple(self.x_coupling, "x") + self._couple(self.y_coupling, "y") + super()._sync()
--- a/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
+++ b/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
@@ -17,7 +17,6 @@ Additional reference:
 """
 
 import numpy as np
-import symengine as se
 from jitcdde import input as system_input
 
 from ..builder.base.network import Network, Node
@@ -59,14 +58,15 @@ class FitzHughNagumoMass(NeuralMass):
     ]
     required_couplings = ["network_x", "network_y"]
 
-    def __init__(self, params=None):
-        super().__init__(params=params or DEFAULT_PARAMS)
+    def __init__(self, params=None, seed=None):
+        super().__init__(params=params or DEFAULT_PARAMS, seed=seed)
 
     def _initialize_state_vector(self):
         """
         Initialize state vector.
         """
-        self.initial_state = [0.05 * np.random.uniform(0, 1)] * self.num_state_variables
+        np.random.seed(self.seed)
+        self.initial_state = (0.05 * np.random.uniform(-1, 1, size=(self.num_state_variables,))).tolist()
 
     def _derivatives(self, coupling_variables):
         [x, y] = self._unwrap_state_vector()
@@ -103,12 +103,14 @@ class FitzHughNagumoNetworkNode(Node):
     default_network_coupling = {"network_x": 0.0, "network_y": 0.0}
     default_output = "x"
 
-    def __init__(self, params=None):
+    def __init__(self, params=None, seed=None):
         """
         :param params: parameters of the FitzHugh-Nagumo mass
         :type params: dict|None
+        :param seed: seed for random number generator
+        :type seed: int|None
         """
-        fhn_mass = FitzHughNagumoMass(params)
+        fhn_mass = FitzHughNagumoMass(params, seed=seed)
         fhn_mass.index = 0
         super().__init__(neural_masses=[fhn_mass])
 
@@ -127,7 +129,7 @@ class FitzHughNagumoNetwork(Network):
     sync_variables = ["network_x", "network_y"]
 
     def __init__(
-        self, connectivity_matrix, delay_matrix, mass_params=None, x_coupling="diffusive", y_coupling="none",
+        self, connectivity_matrix, delay_matrix, mass_params=None, x_coupling="diffusive", y_coupling="none", seed=None,
     ):
         """
         :param connectivity_matrix: connectivity matrix for between nodes
@@ -147,12 +149,15 @@ class FitzHughNagumoNetwork(Network):
         :param y_coupling: how to couple `y` variables in the nodes,
             "diffusive", "additive", or "none"
         :type y_coupling: str
+        :param seed: seed for random number generator
+        :type seed: int|None
         """
         mass_params = self._prepare_mass_params(mass_params, connectivity_matrix.shape[0])
+        seeds = self._prepare_mass_params(seed, connectivity_matrix.shape[0], native_type=int)
 
         nodes = []
         for i, node_params in enumerate(mass_params):
-            node = FitzHughNagumoNetworkNode(params=node_params)
+            node = FitzHughNagumoNetworkNode(params=node_params, seed=seeds[i])
             node.index = i
             node.idx_state_var = i * node.num_state_variables
             nodes.append(node)
@@ -169,29 +174,6 @@ class FitzHughNagumoNetwork(Network):
 
         self.x_coupling = x_coupling
         self.y_coupling = y_coupling
-
-    def _couple(self, coupling_type, coupling_variable):
-        assert coupling_variable in self.coupling_symbols
-        if coupling_type == "additive":
-            return self._additive_coupling(self.coupling_symbols[coupling_variable], f"network_{coupling_variable}",)
-        elif coupling_type == "diffusive":
-            return self._diffusive_coupling(self.coupling_symbols[coupling_variable], f"network_{coupling_variable}",)
-        elif coupling_type == "none":
-            return self._no_coupling(f"network_{coupling_variable}")
-        else:
-            raise ValueError(f"Unknown coupling type: {coupling_type}")
-
-    def init_network(self):
-        # create symbol for each node for input
-        self.sync_symbols = {
-            f"{symbol}_{node_idx}": se.Symbol(symbol)
-            for symbol in self.sync_variables
-            for node_idx in range(self.num_nodes)
-        }
-        for node_idx, node in enumerate(self.nodes):
-            node.init_node(start_idx_for_noise=node_idx * node.num_noise_variables)
-        assert all(node.initialised for node in self)
-        self.initialised = True
 
     def _sync(self):
         return self._couple(self.x_coupling, "x") + self._couple(self.y_coupling, "y") + super()._sync()
--- a/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
+++ b/neurolib/models/multimodel/builder/fitzhugh_nagumo.py
@@ -47,7 +47,7 @@ class FitzHughNagumoMass(NeuralMass):
     num_noise_variables = 2
     coupling_variables = {0: "x", 1: "y"}
     state_variable_names = ["x", "y"]
-    required_parameters = [
+    required_params = [
         "alpha",
         "beta",
         "gamma",
@@ -59,8 +59,8 @@ class FitzHughNagumoMass(NeuralMass):
     ]
     required_couplings = ["network_x", "network_y"]
 
-    def __init__(self, parameters=None):
-        super().__init__(parameters=parameters or DEFAULT_PARAMS)
+    def __init__(self, params=None):
+        super().__init__(params=params or DEFAULT_PARAMS)
 
     def _initialize_state_vector(self):
         """
@@ -72,20 +72,20 @@ class FitzHughNagumoMass(NeuralMass):
         [x, y] = self._unwrap_state_vector()
 
         d_x = (
-            -self.parameters["alpha"] * x ** 3
-            + self.parameters["beta"] * x ** 2
-            + self.parameters["gamma"] * x
+            -self.params["alpha"] * x ** 3
+            + self.params["beta"] * x ** 2
+            + self.params["gamma"] * x
             - y
             + coupling_variables["network_x"]
             + system_input(self.noise_input_idx[0])
-            + self.parameters["ext_input_x"]
+            + self.params["ext_input_x"]
         )
 
         d_y = (
-            (x - self.parameters["delta"] - self.parameters["epsilon"] * y) / self.parameters["tau"]
+            (x - self.params["delta"] - self.params["epsilon"] * y) / self.params["tau"]
             + coupling_variables["network_y"]
             + system_input(self.noise_input_idx[1])
-            + self.parameters["ext_input_y"]
+            + self.params["ext_input_y"]
         )
 
         return [d_x, d_y]
@@ -103,12 +103,12 @@ class FitzHughNagumoNetworkNode(Node):
     default_network_coupling = {"network_x": 0.0, "network_y": 0.0}
     default_output = "x"
 
-    def __init__(self, parameters=None):
+    def __init__(self, params=None):
         """
-        :param parameters: parameters of the FitzHugh-Nagumo mass
-        :type parameters: dict|None
+        :param params: parameters of the FitzHugh-Nagumo mass
+        :type params: dict|None
         """
-        fhn_mass = FitzHughNagumoMass(parameters)
+        fhn_mass = FitzHughNagumoMass(params)
         fhn_mass.index = 0
         super().__init__(neural_masses=[fhn_mass])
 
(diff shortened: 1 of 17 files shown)


neurolib/models/wc/timeIntegration.py
```diff
@@ -232,7 +232,7 @@ def S_I(x):
                     * S_I(
                         c_excinh * excs[no, i - 1]  # input from the excitatory population
                         - c_inhinh * inhs[no, i - 1]  # input from within the inhibitory population
-                        + exc_input_d[no]  # input from other nodes
+                        + inh_input_d[no]  # input from other nodes
```
jajcayn, line 235: I believe there is a bug in neurolib's native Wilson-Cowan so I fixed it, can @ChristophMetzner confirm? (excitatory input / coupling was given to inhibitory mass)

tests/multimodel/test_fitzhugh_nagumo.py
```diff
@@ -0,0 +1,176 @@
...
+        self.assertTrue(isinstance(fhn, FitzHughNagumoNetworkNode))
+        self.assertEqual(len(fhn), 1)
+        self.assertDictEqual(fhn[0].params, DEFAULT_PARAMS)
+        self.assertEqual(len(fhn.default_network_coupling), 2)
+        np.testing.assert_equal(np.array(fhn[0].initial_state), fhn.initial_state)
+
+    def test_run(self):
+        fhn = self._create_node()
+        all_results = []
+        for backend, noise_func in BACKENDS_TO_TEST.items():
```
jajcayn, line 86: here runs from two backends are compared
and similarly in other tests

tests/multimodel/test_fitzhugh_nagumo.py
```diff
@@ -0,0 +1,176 @@
...
+                np.vstack([result[state_var].values.flatten().astype(float) for result in all_results])
+            )
+            self.assertTrue(np.greater(corr_mat, CORR_THRESHOLD).all())
+
+    def test_compare_w_neurolib_native_model(self):
+        """
+        Compare with neurolib's native FitzHugh-Nagumo model.
+        """
+        # run this model
+        fhn_multi = self._create_node()
```
jajcayn, line 109: here run against native `neurolib`'s implementation is tested (if available, hence not for Wong-Wang)

tests/multimodel/test_wong_wang.py
```diff
@@ -0,0 +1,184 @@
...
+        self.assertEqual(len(rww.noise_input_idx), rww.num_noise_variables)
+
+    def test_run(self):
+        rww = self._create_mass()
+        result = self._run_mass(rww, DURATION, DT)
+        self.assertTrue(isinstance(result, np.ndarray))
+        self.assertTupleEqual(result.shape, (int(DURATION / DT), rww.num_state_variables))
+
+
+# class TestWongWangNetworkNode(unittest.TestCase):
```
jajcayn, line 105: will finish once tested against TVB's implementation of Reduced Wong-Wang

neurolib/models/wc/timeIntegration.py
ChristophMetzner: Wait, that is supposed to be excitatory input, because there is no inhibition between different nodes. However, we can discuss whether the inhibitory population should get any input from other nodes or if it should only be excitatory to excitatory coupling between nodes.

jajcayn: Ah, so the model is expected to have excitatory coupling at the network level also in inhibitory mass? then I am sorry, I thought it's a bug:) I'll remove then

jajcayn: fixed (I mean taken back)... in light of this I also added excitatory -> inhibitory coupling to multimodel's implementation :)

ChristophMetzner: I guess most models only have excitatory-excitatory coupling at the network level, but excitatory-inhibitory coupling also exists in biology, so it does make some sense to have this coupling. However, I am also fine with only having excitatory-excitatory coupling in the WC to keep it consistent with all other models.

jajcayn: I totally agree - there are exc->inh coupling in the biology and e.g. I am using exc-> inh at the network level to the thalamus... I left native WC in `neurolib` as is (that is **including** the coupling) and for multimodel implementation, I added it as well, but it can be "turned off" with one line, so you can use both versions and even compare them etc.

ChristophMetzner: Awesome!

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
```diff
@@ -0,0 +1,179 @@
...
+            (x - self.params["delta"] - self.params["epsilon"] * y) / self.params["tau"]
+            + coupling_variables["network_y"]
+            + system_input(self.noise_input_idx[1])
+            + self.params["ext_input_y"]
+        )
+
+        return [d_x, d_y]
+
+
+class FitzHughNagumoNetworkNode(Node):
```
caglorithm, line 94:
```suggestion
class FitzHughNagumoNode(Node):
```
To avoid confusion with `Network`

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
```diff
@@ -0,0 +1,179 @@
...
+        :type y_coupling: str
+        :param seed: seed for random number generator
+        :type seed: int|None
+        """
+        mass_params = self._prepare_mass_params(mass_params, connectivity_matrix.shape[0])
+        seeds = self._prepare_mass_params(seed, connectivity_matrix.shape[0], native_type=int)
+
+        nodes = []
+        for i, node_params in enumerate(mass_params):
+            node = FitzHughNagumoNetworkNode(params=node_params, seed=seeds[i])
```
caglorithm, line 160:
```suggestion
            node = FitzHughNagumoNode(params=node_params, seed=seeds[i])
```

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
```diff
@@ -0,0 +1,179 @@
...
+        "alpha",
+        "beta",
+        "gamma",
+        "delta",
+        "epsilon",
+        "tau",
+        "ext_input_x",
+        "ext_input_y",
+    ]
+    required_couplings = ["network_x", "network_y"]
```
caglorithm, line 59: The "coupling type" `x_coupling="diffusive", y_coupling="none"` is an argument of `FitzHughNagumoNetwork`. 

Couldn't it be defined here and stay fixed for the model? It could be easier to automate the coupling then in the superclass (see my comment at end of file in `_synch()`).

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
```diff
@@ -0,0 +1,179 @@
...
+        # assert they are the same
+        assert all(all_couplings[0] == coupling for coupling in all_couplings)
+        # invert as to name: idx
+        self.coupling_symbols = {v: k for k, v in all_couplings[0].items()}
+
+        self.x_coupling = x_coupling
+        self.y_coupling = y_coupling
+
+    def _sync(self):
+        return self._couple(self.x_coupling, "x") + self._couple(self.y_coupling, "y") + super()._sync()
```
caglorithm, line 179: Couldn't this be automated in the superclass actually? 
```python
return np.sum([<look upver couplings here>], super().sync())
```

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        # get all coupling variables
+        all_couplings = [mass.coupling_variables for node in self.nodes for mass in node.masses]
+        # assert they are the same
+        assert all(all_couplings[0] == coupling for coupling in all_couplings)
+        # invert as to name: idx
+        self.coupling_symbols = {v: k for k, v in all_couplings[0].items()}
+
+        self.x_coupling = x_coupling
+        self.y_coupling = y_coupling
```
caglorithm, line 176: See next comment, could maybe be refactored and user-agnostic.

neurolib/models/multimodel/builder/thalamus.py
```diff
@@ -0,0 +1,442 @@
...
+        return (
+            self.params["g_T"]
+            * self._m_inf_T(voltage)
+            * self._m_inf_T(voltage)
+            * h_T_value
+            * (voltage - self.params["E_Ca"])
+        )
+
+
+class ThalamocorticalPopulation(ThalamicMass):
```
caglorithm, line 115:
```suggestion
class ThalamocorticalMass(ThalamicMass):
```
I suggest, we use either "Mass" or "Population" all the way (and avoid using the other). We've said maybe "Population" is nicer and more general, but I think both is good! You decide?

neurolib/models/multimodel/builder/thalamus.py
```diff
@@ -0,0 +1,442 @@
...
+            - 2 * self.params["gamma_e"] * dsyn_ext
+        )
+        d_dsyn_inh = (
+            self.params["gamma_r"] ** 2 * (coupling_variables["node_exc_inh"] - syn_inh)
+            - 2 * self.params["gamma_r"] * dsyn_inh
+        )
+        # firing rate as dummy dynamical variable with infinitely fast
+        # fixed-point dynamics
+        firing_rate_now = self._get_firing_rate(voltage)
+        d_firing_rate = -self.params["lambda"] * (firing_rate - firing_rate_now)
```
caglorithm, line 276: These equations are... a lot! Holy moly.

neurolib/models/multimodel/builder/thalamus.py
```diff
@@ -0,0 +1,442 @@
...
+        super().__init__(params=params or DEFAULT_PARAMS_TRN)
+
+    def _m_inf_T(self, voltage):
+        return 1.0 / (1.0 + exp(-(voltage + 52.0) / 7.4))
+
+    def _h_inf_T(self, voltage):
+        return 1.0 / (1.0 + exp((voltage + 80.0) / 5.0))
+
+    def _tau_h_T(self, voltage):
+        return (85.0 + 1.0 / (exp((voltage + 48.0) / 4.0) + exp(-(voltage + 407.0) / 50.0))) / 3.7371928
```
caglorithm, line 348: Why are hard-coded constants here?

neurolib/models/multimodel/builder/thalamus.py
```diff
@@ -0,0 +1,442 @@
...
+        :param tcr_params: parameters for the excitatory (TCR) mass
+        :type tcr_params: dict|None
+        :param trn_params: parameters for the inhibitory (TRN) mass
+        :type trn_params: dict|None
+        :param connectivity: local connectivity matrix
+        :type connectivity: np.ndarray
+        """
+        tcr_mass = ThalamocorticalPopulation(params=tcr_params)
+        tcr_mass.index = 0
+        trn_mass = ThalamicReticularPopulation(params=trn_params)
```
caglorithm, line 435:
```suggestion
        tcr_mass = ThalamocorticalMass(params=tcr_params)
        tcr_mass.index = 0
        trn_mass = ThalamicReticularMass(params=trn_params)
```

neurolib/models/multimodel/builder/thalamus.py
```diff
@@ -0,0 +1,442 @@
...
+            d_m_h2,
+            d_syn_ext,
+            d_syn_inh,
+            d_dsyn_ext,
+            d_dsyn_inh,
+            d_firing_rate,
+        ]
+
+
+class ThalamicReticularPopulation(ThalamicMass):
```
caglorithm, line 292:
```suggestion
class ThalamicReticularMass(ThalamicMass):
```
see comment upstairs :)

neurolib/models/multimodel/builder/wilson_cowan.py
```diff
@@ -0,0 +1,240 @@
...
+        assert len(self.sync_variables) == 2
+
+    def _sync(self):
+        # excitatory population within the node is first, hence the
+        # within_node_idx is 0
+        return (
+            self._additive_coupling(within_node_idx=0, symbol="network_exc_exc")
+            + self._additive_coupling(within_node_idx=0, symbol="network_inh_exc")
+            + super()._sync()
+        )
```
caglorithm, line 240: Like in other models, maybe this can be refactored.

tests/multimodel/base/test_backend.py
```diff
@@ -178,6 +183,21 @@ def test_run_jitcdde(self):
         self.assertTrue(all(dim in results.dims for dim in ["time", "node"]))
         self.assertDictEqual(results.attrs, self.EXTRA_ATTRS)
 
+    def test_jitcdde_other_features(self):
+        system = BackendTestingHelper()
+        _ = system.run(self.DURATION, self.DT, ZeroInput(self.DURATION, self.DT).as_cubic_splines(), backend="jitcdde")
+        system.backend_instance._check()
+        system.backend_instance.dde_system.reset_integrator()
+        system.backend_instance._integrate_blindly(system.max_delay)
+        system.clean()
```
caglorithm, line 192: 😎

tests/multimodel/test_fitzhugh_nagumo.py
```diff
@@ -0,0 +1,176 @@
...
+            self.assertTrue(
+                all(result[state_var].shape == (int(DURATION / DT), 1) for state_var in fhn.state_variable_names[0])
+            )
+            all_results.append(result)
+        # test results are the same from different backends
+        for state_var in all_results[0]:
+            corr_mat = np.corrcoef(
+                np.vstack([result[state_var].values.flatten().astype(float) for result in all_results])
+            )
+            self.assertTrue(np.greater(corr_mat, CORR_THRESHOLD).all())
```
caglorithm, line 102: 👍🏽

tests/multimodel/test_fitzhugh_nagumo.py
caglorithm: TODO: Implement WW soon :)

caglorithm: Ehm, have you tested it with noise actually?

tests/multimodel/test_hopf.py
```diff
@@ -146,24 +146,23 @@ def test_compare_w_neurolib_native_model(self):
...
         # delays <-> length matrix
-        fhn_neurolib.params["signalV"] = 1.0
-        fhn_neurolib.params["coupling"] = "diffusive"
-        fhn_neurolib.params["sigma_ou"] = 0.0
-        fhn_neurolib.params["xs_init"] = fhn_multi.initial_state[::2][:, np.newaxis]
-        fhn_neurolib.params["ys_init"] = fhn_multi.initial_state[1::2][:, np.newaxis]
-        fhn_neurolib.run()
+        hopf_neurolib.params["signalV"] = 1.0
+        hopf_neurolib.params["coupling"] = "diffusive"
+        hopf_neurolib.params["sigma_ou"] = 0.0
```
caglorithm, line 160: I'm guessing `params["sigma_ou"] > 0.0`  is nontrivial...

tests/multimodel/test_wilson_cowan.py
```diff
@@ -0,0 +1,196 @@
...
+        # test results are the same from different backends
+        for state_var in all_results[0]:
+            corr_mat = np.corrcoef(
+                np.vstack([result[state_var].values.flatten().astype(float) for result in all_results])
+            )
+            self.assertTrue(np.greater(corr_mat, CORR_THRESHOLD).all())
+
+    def test_compare_w_neurolib_native_model(self):
+        """
+        Compare with neurolib's native Wilson-Cowan model.
```
caglorithm, line 170:
```suggestion
        Compare with neurolib's native Thalamic model.
```

jajcayn:
rebased for now, will address review later:)

neurolib/models/multimodel/builder/thalamus.py
jajcayn: I'd go with `Mass` then :)

neurolib/models/multimodel/builder/thalamus.py
jajcayn: well, the same hardcoded constants are in the original paper and also in the papers that the original paper (Costa et al.) refers to... those are from the experiments in vitro done by Destexhe back in '93 or so... maybe I am not a huge fan of that but in thalamic modelling people use a lot of hard-coded constants, in particular for channel dynamics of a sigmoid type...

tests/multimodel/test_hopf.py
jajcayn: I haven't with noise so far... too nontrivial with all the seeds lying around, I guess you never have the same noise realisations without hacking deep into the framework...
I just did tests with noise in notebooks - e.g. compared ALN run with OU from `neurolib`'s native and from builder... I have some notebooks planned for multimodels - I wanted to do noise-comparison of at least ALN and maybe FHN.. are you ok with this being just in the notebook for visual inspection and not in the rigorous test?

tests/multimodel/test_wilson_cowan.py
jajcayn: ?

neurolib/models/multimodel/builder/fitzhugh_nagumo.py
jajcayn: I moved this part to `Network` base class and create a new way of how couplings are determined for the basic models

tests/multimodel/test_wilson_cowan.py
caglorithm: mhhh sorry I must've mixed it up :)

tests/multimodel/test_hopf.py
caglorithm: I'm ok with it, having the exact same noise vectors is probably tricky, yes.... it could be easier for fhn which has only one mass, e.g. one noise vector. in neurolib right now the noise vector is generated before random parameters afaik. which means that setting the seed in multimodel just before generating the noise vector (of same length) should yield the same noise.....

PR #97, opened 2020-07-03
MultiModel PR3

jajcayn:
Finally, the holy grail! ~AdEx~ ALN model!

so yes, only ~AdEx~ ALN model and its Node and Network, of course, bunch of tests...
And yes - need to ignore `adex.py` in `codecov` because of `numba` and bumped the version to 0.5.7

I'll add some comments directly into the code as a review.
Final PR, then the actual fun begins: connect all this to `neurolib`'s core!


neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+}
+# matrices as [to, from], masses as (EXC, INH)
+# EXC is index 0, INH is index 1
+DEFAULT_ADEX_NODE_CONNECTIVITY = np.array([[0.3, 0.5], [0.3, 0.5]])
+# same but delays, in ms
+DEFAULT_ADEX_NODE_DELAYS = np.array([[4.0, 2.0], [4.0, 2.0]])
+
+
+@numba.njit()
+def _get_interpolation_values(xi, yi, sigma_range, mu_range, d_sigma, d_mu):
```
jajcayn, line 96: `numba` function for table lookup - this just gets the correct table indices

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        xid1 = -1
+        dxid = 0.0
+        yid = (yi - mu_range[0]) / d_mu
+        yid1 = np.floor(yid)
+        dyid = yid - yid1
+        return int(xid1), int(yid1), dxid, dyid
+
+
+@numba.njit()
+def _table_lookup(
```
jajcayn, line 179: `numba` function for the actual lookup

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    name = "AdEx mean-field mass"
+    label = "AdExMass"
+    # define python callback function name for table lookup (linear-nonlinear
+    # approximation of Fokker-Planck equation)
+    python_callbacks = ["firing_rate_lookup", "voltage_lookup", "tau_lookup"]
+
+    num_noise_variables = 1
+
+    @staticmethod
+    def _rescale_strengths(params):
```
jajcayn, line 213: ALN uses rescaled strengths so it's done here... plus `tau_m` is computed from other parameters

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        logging.info("Loading done, all quantities loaded.")
+        # close the file
+        loaded_h5.close()
+        self.lin_nonlin_fname = filename
+
+    def update_params(self, params_dict):
+        """
+        Update parameters as in the base class but also rescale.
+        """
+        # if we are changing C_m or g_L, update tau_m as well
```
jajcayn, line 264: because of the rescaling it's messy to update parameters, but should be readable... practically all parameters are updated using `dict.update()` so here I'm just checking whether parameters from which the rescaling is done are also updated and if so, rescaling is done again

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        # update all parameters finally
+        super().update_params(params_dict)
+
+    def describe(self):
+        return {
+            **super().describe(),
+            "lin_nonlin_cascade_filename": self.lin_nonlin_fname,
+        }
+
+    def _callbacks(self):
```
jajcayn, line 291: now the famous callbacks:
for `jitcdde` backend it was tricky - it compiles symbolic code into C and then integrates - not doable with table lookup. the author of `jitcdde` was so nice, he implemented a functionality called `python_callbacks` (functions which you can call in derivatives definition)
the definition of callbacks is similar as for `sync` variables hence it must be a list of tuples, where each tuple is callback name (as `symengine.Symbol`) and then its definition as a python function, plus number of parameters (hence tuple of length 3).. of course, for speed callbacks are defined here as `numba` jitted functions

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        """
+        callbacks_list = [
+            (self.callback_functions["firing_rate_lookup"], self.firing_rate_lookup, 2),
+            (self.callback_functions["voltage_lookup"], self.voltage_lookup, 2),
+            (self.callback_functions["tau_lookup"], self.tau_lookup, 2),
+        ]
+        self._validate_callbacks(callbacks_list)
+        return callbacks_list
+
+    def _numba_callbacks(self):
```
jajcayn, line 303: the tricky part then was - callbacks for `numba` backend had to be defined in some other way! this is because callbacks in `jitcdde` can be class methods, but for `numba` backend they cannot - if you wanna use function in `numba` jitted function, that function must be also jitted! So functionally, of course, the `jitcdde` and `numba` callbacks are the same, but they need to be shipped to the backend differently...

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        """
+        Define numba callbacks - has to be different than jitcdde callbacks
+        because of the internals.
+        """
+
+        def _table_numba_gen(sigma_range, mu_range, d_sigma, d_mu, cascade):
```
jajcayn, line 309: this is a little trick where instead of something like
```python
partial(_table_lookup, sigma_range=..., mu_range=..., ...)
```
which cannot be wrapped in `numba.njit()` you do nested function and when you call it, you can still use `numba.njit()`... so this is really just `partial()` with the possibility of using `numba.njit()` lates

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+
+            def inner(current_mu, current_sigma):
+                return _table_lookup(current_mu, current_sigma, sigma_range, mu_range, d_sigma, d_mu, cascade,)
+
+            return inner
+
+        return [
+            (
+                "firing_rate_lookup",
+                numba.njit(
```
jajcayn, line 325: here wrapping `partial()` functions with `numba.njit()`, which, again, can be done thanks to that inner function trick

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+            ),
+            (
+                "tau_lookup",
+                numba.njit(
+                    _table_numba_gen(self.sigma_range, self.mu_range, self.d_sigma, self.d_mu, self.tau_cascade)
+                ),
+            ),
+        ]
+
+    def firing_rate_lookup(self, y, current_mu, current_sigma):
```
jajcayn, line 345: this is the actual definition of first callback function - table lookup for firing rate

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+            current_mu,
+            current_sigma,
+            self.sigma_range,
+            self.mu_range,
+            self.d_sigma,
+            self.d_mu,
+            self.firing_rate_cascade,
+        )
+
+    def voltage_lookup(self, y, current_mu, current_sigma):
```
jajcayn, line 360: this is the actual definition of second callback function - table lookup for voltage

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        """
+        Translate mean and std. deviation of the current to voltage using
+        linear-nonlinear lookup table for AdEx.
+        """
+        return _table_lookup(
+            current_mu, current_sigma, self.sigma_range, self.mu_range, self.d_sigma, self.d_mu, self.voltage_cascade,
+        )
+
+    def tau_lookup(self, y, current_mu, current_sigma):
```
jajcayn, line 369: this is the actual definition of third callback function - table lookup for tau

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    def _initialize_state_vector(self):
+        """
+        Initialize state vector.
+        """
+        np.random.seed(self.seed)
+        self.initial_state = (
+            np.random.uniform(0, 1, self.num_state_variables) * np.array([3.0, 200.0, 0.5, 0.5, 0.001, 0.001, 0.01])
+        ).tolist()
+
+    def _compute_couplings(self, coupling_variables):
```
jajcayn, line 441: dynamics definition for AdEx is kinda ugly and long so I defined some helper convenience functions such as this - this just computes coupling - both exc -> exc and inh->exc and their squared variants

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    def _initialize_state_vector(self):
+        """
+        Initialize state vector.
+        """
+        np.random.seed(self.seed)
+        self.initial_state = (
+            np.random.uniform(0, 1, self.num_state_variables) * np.array([3.0, 0.5, 0.5, 0.01, 0.01, 0.01])
+        ).tolist()
+
+    def _compute_couplings(self, coupling_variables):
```
jajcayn, line 599: dynamics definition for AdEx is kinda ugly and long so I defined some helper convenience functions such as this - this just computes coupling - both inh -> inh and exc-> inh and their squared variants

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    ]
+
+    default_network_coupling = {
+        "network_exc_exc": 0.0,
+        "network_exc_exc_sq": 0.0,
+    }
+
+    default_output = f"q_mean_{EXC}"
+
+    def _rescale_connectivity(self):
```
jajcayn, line 709: and again - the local E-I connectivity in ALN model is rescaled... so rescaling here

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        inhibitory_mass = InhibitoryAdExMass(
+            params=inh_params, lin_nonlin_cascade_filename=inh_lin_nonlin_cascade_filename, seed=inh_seed
+        )
+        inhibitory_mass.index = 1
+        super().__init__(
+            neural_masses=[excitatory_mass, inhibitory_mass], local_connectivity=connectivity, local_delays=delays,
+        )
+        self._rescale_connectivity()
+
+    def update_params(self, params_dict):
```
jajcayn, line 778: as with individual masses - when updating parameters we need to check whether the connectivity itself is changed and if so - it has to be rescaled again

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+                self.sync_symbols[f"node_exc_inh_sq_{self.index}"],
+                sum([connectivity_sq[row, col] for row in self.excitatory_masses for col in self.inhibitory_masses]),
+            ),
+            (
+                # inh -> inh squared connectivity
+                self.sync_symbols[f"node_inh_inh_sq_{self.index}"],
+                sum([connectivity_sq[row, col] for row in self.inhibitory_masses for col in self.inhibitory_masses]),
+            ),
+        ]
+        return super()._sync() + sq_connectivity
```
jajcayn, line 815: `super()._sync()` takes care of not-squared variants of the coupling

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        if rescale_flag:
+            self._rescale_connectivity()
+
+    def _sync(self):
+        """
+        Apart from basic EXC<->INH connectivity, construct also squared
+        variants.
+        """
+        connectivity_sq = self.connectivity ** 2 * self.inputs
+        sq_connectivity = [
```
jajcayn, line 793: need to define squared variants of the coupling

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+class AdExNetwork(Network):
+    """
+    Whole brain network of adaptive exponential integrate-and-fire mean-field
+    excitatory and inhibitory nodes.
+    """
+
+    name = "AdEx mean-field network"
+    label = "AdExNet"
+
+    sync_variables = ["network_exc_exc", "network_exc_exc_sq"]
```
jajcayn, line 827: just exc->exc in the network (not as in Wilson-Cowan or Wong-Wang), but also a squared variant...

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+                mass.noise_input_idx = [2 * i + mass.index]
+            nodes.append(node)
+
+        super().__init__(
+            nodes=nodes, connectivity_matrix=connectivity_matrix, delay_matrix=delay_matrix,
+        )
+        # assert we have 2 sync variable
+        assert len(self.sync_variables) == 2
+
+    def _sync(self):
```
jajcayn, line 924: unfortunately - due to squared coupling it's just easier to redefine `network._sync()` rather than trying to accommodate all version - then the superclass `Network` would be too complicated... so I'm redefining here to accommodate network couplings - both basic and squared variants are computed here

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+            # regular additive coupling
+            self._additive_coupling(within_node_idx=coupling_var_idx, symbol="network_exc_exc")
+            # additive coupling with squared weights
+            + self._additive_coupling(
+                within_node_idx=coupling_var_idx,
+                symbol="network_exc_exc_sq",
+                # multiplier is connectivity again, then they'll be squared
+                connectivity_multiplier=self.connectivity,
+            )
+            + super()._sync()
```
jajcayn, line 943: `super()._sync()` takes care of the sum of nodal couplings

jajcayn:
I hope I finally resolved all failing tests... the problem was that sometimes (only on Linux 🤷) model run produced NaNs and hence the correlation when testing run results was also nan... now I am checking for nans and testing only when a model produced actual numbers...

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    Low-dimensional spike rate models derived from networks of adaptive
+    integrate-and-fire neurons: comparison and implementation. PLoS Comput Biol,
+    13(6), e1005545.
+
+Additional reference:
+    Cakan C., Obermayer K. (2020). Biophysically grounded mean-field models of
+    neural populations under electrical stimulation. PLoS Comput Biol, 16(4),
+    e1007822.
+"""
+
```
caglorithm, line 15:
```suggestion
```
I would remove this (like in all other model files).

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+from h5py import File
+from jitcdde import input as system_input
+
+from ..builder.base.constants import EXC, INH, LAMBDA_SPEED
+from ..builder.base.network import Network, SingleCouplingExcitatoryInhibitoryNode
+from ..builder.base.neural_mass import NeuralMass
+
+DEFAULT_CASCADE_FILENAME = "quantities_cascade.h5"
+
+DEFAULT_PARAMS_EXC = {
```
caglorithm, line 32: A lot of these parameters have different names than in the original `aln` model. This could be pretty confusing, no? Why not load parameters from `loadDefaultParams.py`?

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+    "C_m": 200.0,
+    "g_L": 10.0,
+    # external drives
+    "ext_current": 0.0,
+    "ext_rate": 0.0,
+    "lambda": LAMBDA_SPEED,
+}
+# matrices as [to, from], masses as (EXC, INH)
+# EXC is index 0, INH is index 1
+DEFAULT_ADEX_NODE_CONNECTIVITY = np.array([[0.3, 0.5], [0.3, 0.5]])
```
caglorithm, line 90: Where exactly do these parameters come from? (What are their analoge in `aln`?) Are these defined by the `c`'s? Maybe substitute?

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+        """
+        Update parameters as in the base class but also rescale.
+        """
+        # if we are changing C_m or g_L, update tau_m as well
+        if any(k in params_dict for k in ("C_m", "g_L")):
+            C_m = params_dict["C_m"] if "C_m" in params_dict else self.params["C_m"]
+            g_L = params_dict["g_L"] if "g_L" in params_dict else self.params["g_L"]
+            params_dict["tau_m"] = C_m / g_L
+
+        # if we are chaning any of the J_exc_max, tau_syn_exc or c_global, rescale c_global
```
caglorithm, line 270:
```suggestion
        # if we are changing any of the J_exc_max, tau_syn_exc or c_global, rescale c_global
```

neurolib/models/multimodel/builder/adex.py
caglorithm: ok 🤯

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+            + (exc_inp_sq - 2.0 * self.params["tau_syn_exc"] * (exc_inp + 1.0)) * I_syn_sigma_exc
+        ) / (self.params["tau_syn_exc"] ** 2)
+
+        d_I_syn_sigma_inh = (
+            (1.0 - I_syn_mu_inh) ** 2 * inh_inp_sq
+            + (inh_inp_sq - 2.0 * self.params["tau_syn_inh"] * (inh_inp + 1.0)) * I_syn_sigma_inh
+        ) / (self.params["tau_syn_inh"] ** 2)
+        # firing rate as dummy dynamical variable with infinitely fast
+        # fixed-point dynamics
+        d_firing_rate = -self.params["lambda"] * (firing_rate - firing_rate_now)
```
caglorithm, line 669: `LAMBDA_SPEED = 10.0`, it's not "infinitely fast" is it?

neurolib/models/multimodel/builder/adex.py
```diff
@@ -0,0 +1,944 @@
...
+            # taus are constant per col and depends only on "from" mass
+            tau_mat[:, col] = mass_from.params[f"tau_syn_{mass_from.mass_type.lower()}"]
+            # Js are specific: take J from "to" mass but of type "from" mass
+            for row, mass_to in enumerate(self.masses):
+                J_mat[row, col] = mass_to.params[f"J_{mass_from.mass_type.lower()}_max"]
+
+        # multiplication with tau makes the increase of synaptic activity
+        # subject to a single input spike invariant to tau and division by J
+        # ensures that mu = J*s will result in a PSP of exactly c for a single
+        # spike
```
caglorithm, line 727: 😎

caglorithm (commented):
Beautiful work! Took me quite a while, sorry for that!

neurolib/models/multimodel/builder/adex.py
jajcayn: no, it's not. BUT with typical timescale as we model with (dt=0.1 or 0.01) it's more than enough.. with higher lambdas the model sometimes blow-up.. and with `lambda_speed = 10.0` the results are comparable as if not using this trick

neurolib/models/multimodel/builder/adex.py
jajcayn: yes, these are the `c`s in the original ALN model. Since the whole design of `MultiModel` is such that we have individual masses, then nodes and then network this is the connectivity matrix for AdEx Node (with one excitatory and one inhibitory mass).

As per names - I will open an issue for this and personally, I would deal with it in a separate PR (as well as with other `MultiModel` models)

neurolib/models/multimodel/builder/adex.py
caglorithm: MHhh did you test this? What is the unit of this equation? ms? We can use this IMO but we should make sure that we don't introduce artifacts when using the AdEx like this... maybe compare bifurcation diagrams later some time. Anyways, I think it should be thoroughly tested before we recommend the use of this. What do you think?

neurolib/models/multimodel/builder/adex.py
jajcayn: the set of parameters in `loadDefaultParams.py` is just too different. The parameters in the original ALN model implementation contains all parameters even for the connectivity, structural matrix, all the delays, input noise parameters etc. The `MultiModel` design somehow automatically in a natural manner divides the parameter "groups" into parameters for the individual mass, then connectivity / delay between masses when building a node and then connectivity / delays when building a network out of nodes... I feel that loading parameters from this file would be more confusing since then I would need to divide parameters from one dictionary into two dictionaries (one for EXC, one for INH), then make a square matrix out of local connectivity and delays (`c`s and `d`s) etc...

as for the names, I would consolidate all the names in the separate issue for all `MultiModel` models depending on our agreement

neurolib/models/multimodel/builder/adex.py
jajcayn: there are citations for relevant papers in all `MultiModel` model files :) but I can remove them, of course..

neurolib/models/multimodel/builder/adex.py
jajcayn: well, there is a test in `tests/multimodel/test_adex.py` that tests one node and one dummy small network between this implementation and original `neurolib`'s native implementation, meaning it will run both implementations and compare it with Pearson's `r`

but definitely agree; my original plan was after merging all the necessary bits and pieces to work on notebooks that'll thoroughly compare `neurolib`'s native models (ALN, Hopf, WC) with this MultiModel modular implementations. Another set of notebooks would be detailing how to actually use this to build any model of your liking

tests/multimodel/test_adex.py
```diff
@@ -0,0 +1,316 @@
...
+            )
+            all_results.append(result)
+        # test results are the same from different backends
+        for state_var in all_results[0]:
+            corr_mat = np.corrcoef(
+                np.vstack([result[state_var].values.flatten().astype(float) for result in all_results])
+            )
+            self.assertTrue(np.greater(corr_mat, CORR_THRESHOLD).all())
+
+    def test_compare_w_neurolib_native_model(self):
```
jajcayn, line 234: @caglarcakan here is the comparison between native ALN and this one for one node

tests/multimodel/test_adex.py
```diff
@@ -0,0 +1,316 @@
...
+        # test results are the same from different backends
+        for state_var in all_results[0]:
+            all_ts = np.vstack([result[state_var].values.flatten().astype(float) for result in all_results])
+            if np.isnan(all_ts).any():
+                continue
+            corr_mat = np.corrcoef(all_ts)
+            print(corr_mat)
+            self.assertTrue(np.greater(corr_mat, CORR_THRESHOLD).all())
+
+    def test_compare_w_neurolib_native_model(self):
```
jajcayn, line 283: @caglarcakan here is the comparison between native ALN and this one for a small dummy network

neurolib/models/multimodel/builder/adex.py
caglorithm: Sound good! I think this is in an important point we need to keep in mind. The synaptic model of AdEx has time scales of 4 or so... so it's actually less than this one.

neurolib/models/multimodel/builder/adex.py
caglorithm: I actually mean the ones that are already here: 
# number of inputs per neuron from EXC/INH
```
    "K_exc": 800.0,
    "K_inh": 200.0,
    # postsynaptic potential amplitude for global connectome
    "c_global": 0.4,
    # number of incoming EXC connections (to EXC population) from each area
    "K_exc_global": 250.0,
    # synaptic time constant EXC/INH
    "tau_syn_exc": 2.0,  # ms
    "tau_syn_inh": 5.0,  # ms
    # internal Fokker-Planck noise due to random coupling
    "sigma_ext": 1.5,  # mV/sqrt(ms)
    # maximum synaptic current between EXC/INH nodes in mV/ms
    "J_exc_max": 2.43,
    "J_inh_max": -3.3,
```

Many of these are the same as in the `aln` model but have different names.

jajcayn:
Sorry for such a long quiescent period:) I finally renamed all the parameters for all the models within `multimodel` to match those in native `neurolib`. Of course, I'd wait for tests and hope they'd pass...

jajcayn:
will rebase this branch on the master, hang tight!

PR #101, opened 2020-08-14
Aln initial conditions

lenasal:
Reworked initial conditions and parameter input, check input for correct type and shape


caglorithm:
Closing because PR contains a lot of undocumented changes and lots of commented debug code.

PR #105, opened 2020-10-23
Fix: thalamus

jajcayn:
The thalamic model now works with `autochunk`. This was resolved by manually casting all necessary thalamic outputs into 2D.


caglorithm (commented):
LGTM! :)

PR #109, opened 2020-11-07
Introduce sampling dt

jajcayn:
Sampling `dt` now as a parameter of models.

(Do not worry, most lines in this PR are just notebooks:) )


caglorithm:
Hey! Thanks for the code 💪🏾

One issue I see though is the default value of `sampling_dt` – it should be `None` in all cases such that no changes are introduced with existing scripts and pipelines. I think this should be the default behaviour also because I it might be unexpected and I usually don't need or want the signal to be subsampled :)

jajcayn:
yeah, that makes sense actually:) will edit!

caglorithm:
done! lemme know if you think it's ok!

PR #117, opened 2020-12-18
Simplify `jitcdde` backend

jajcayn:
Remove unused options for `jitcdde` backend:
* OpenMP support - `jitcdde` can use OpenMP for compiling symbolic equations into C. However, this comes with tremendous overhead and while I was testing this, OpenMP does not offer almost any speedup
* save / load compiled - there is an option to save compiled C code of the equations and then just load it on next run. In theory, this would significantly speed up computations of large networks where the compilation to C takes a lot of time. However, our code works in a way that parameter values are "hardcoded" into symbolic computation, hence on each parameter change (imagine exploration), we would need to compile again. And therefore I cannot see any use case of this as of now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,11 +5,7 @@ name: ci
 
 on:
   push:
-    branches:
-      - "*"
   pull_request:
-    branches:
-      - "*"
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


caglorithm:
There should be an alternative nobel prize for getting rid of unused code!

jajcayn:
tests passed, all good, merging!

PR #118, opened 2021-01-13
MultiModel small changes & Exploration fix

jajcayn:
- same name for firing rate in `ALN` and `thalamus` models
- better handling of connectivity in `_additive_coupling()` and `_diffusive_coupling()` helpers
- allow `None` in `_construct_input_matrix()` helper as an option for a node to not couple anywhere (useful when the whole network has a different coupling for each node, e.g. thalamus has network `exc -> inh` coupling, but ALN does not)
- some changes in parameter handling:
     - change names in the overall parameter list to mass / node label so you can change all parameter of the all instances of the same type (e.g. `ALN`)
     - update parameters in the noise inputs (for some reason I forgot to add this before...)
- `ParameterSpace` class now have `allow_star_notation` keyword which allows to track little changes in `Exploration` class (e.g. parameters in `pypet` in long-format etc - all because of parameter handling in MultiModel)

### DIFF
--- a/neurolib/models/multimodel/builder/base/neural_mass.py
+++ b/neurolib/models/multimodel/builder/base/neural_mass.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 import numpy as np
 import symengine as se
 from jitcdde import y as state_vector
@@ -83,7 +85,7 @@ class NeuralMass:
         :type seed: int|None
         """
         assert isinstance(params, dict)
-        self.params = params
+        self.params = deepcopy(params)
         self.seed = seed
         # used in determining portion of the full system's state vector
         self.idx_state_var = None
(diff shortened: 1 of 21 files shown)


jajcayn:
@caglorithm ready.. could you check it out, please when you have a moment? little stuff, exploration now works:)

neurolib/models/multimodel/builder/base/neural_mass.py
```diff
@@ -170,9 +175,13 @@ def update_params(self, params_dict):
         :type params_dict: dict
         """
         assert isinstance(params_dict, dict)
-        self.params.update(params_dict)
+        self.params.update({k: v for k, v in params_dict.items() if "noise" not in k})
```
caglorithm, line 178: Could you add a comment here, why this is necessary? User would have to avoid any parameter with label "noise" in it?

caglorithm (commented):
LGTM!

caglorithm (approved):

neurolib/models/multimodel/builder/base/neural_mass.py
jajcayn: yes, because `noise_X` (X is integer) is the key for the noise process that is expected in the mass... I mean, I guess wouldn't have a parameter called `noise`. The only parameter call `noise` would be something like noise variance, but in MultiModel this is done by setting a noise process and connecting it to the population (see `noise_input` attribute of `NeuralMass`)..
so in this function, you first update the parameter dictionary of the mass itself, and then, couple of lines below, you actually set up (update) noise parameters

PR #119, opened 2021-01-18
Exploration results to `xr.DataArray`

jajcayn:
Simple convenience function to return all exploration results to `xr.DataArray`. The model coordinates follow the convention in `model.xr()`, i.e. dimensions "time", "output", and "space". In the MultiModel case, the dimensions added follow star notation, and the attributes then contain the full list of affected parameters.
Has optional `bold: bool` keyword - if True, return DataArray of BOLD model output, otherwise removes BOLD output if simulated and returns default model outputs.

Note: this PR is built on PR #118, hence I am opening this as a draft and will rebase once #118 is merged.


caglorithm:
Nice work! What's the rationale of filtering BOLD (or not filtering it)? Why not put it into `xr`  always?

jajcayn:
so typically BOLD has different `dt`, much slower and single `DataArray` (or `Dataset` for that matter) doesn't support multiple different coordinates
I saw two options:
* put everything into one DataArray with lot of NaNs (where BOLD is not defined), which would be messy and I think not really intuitive
* ask user whether she wants BOLD or default model output - I believe this is more reasonable since you can always ask for both of them and you'll have two nice DataArrays without NaNs

neurolib/optimize/exploration/exploration.py
```diff
@@ -418,6 +421,99 @@ def loadDfResults(self, filename=None, trajectoryName=None):
...
+            )
+            expand_coords = {}
+            for k, v in expl_coords.items():
+                if isinstance(v, (str, float, int)):
+                    expand_coords[k] = [v]
+                elif isinstance(v, np.ndarray):
+                    if np.unique(v).size == 1:
+                        expand_coords[k] = [float(np.unique(v))]
+                    else:
+                        raise ValueError("Cannot squeeze coordinates")
```
caglorithm, line 508: Please explain what the user can do to fix this error.

caglorithm, line 508: This block needs commenting.

tests/test_exploration.py
```diff
@@ -36,6 +37,9 @@ def test_single_node(self):
         search = BoxSearch(model, parameters, filename="test_single_nodes.hdf")
         search.run()
         search.loadResults()
+        dataarray = search.xr()
+        self.assertTrue(isinstance(dataarray, xr.DataArray))
+        self.assertFalse(dataarray.attrs)
```
caglorithm, line 42: ``` self.assertFalse(dataarray.attrs)```
What does this line do?

caglorithm (commented):
> so typically BOLD has different `dt`, much slower and single `DataArray` (or `Dataset` for that matter) doesn't support multiple different coordinates
> I saw two options:
> 
> * put everything into one DataArray with lot of NaNs (where BOLD is not defined), which would be messy and I think not really intuitive
> * ask user whether she wants BOLD or default model output - I believe this is more reasonable since you can always ask for both of them and you'll have two nice DataArrays without NaNs

I see the point of not chosing option 1 and I guess it's fine to have option 2 but it is a bit messy. 

I think most ideally there would be a `utils.xr` which collects all the xr related functions because they are starting to pop up in different parts of the code and I think that's not ideal. The same is true for pandas which is why I have been exporting functions to explrationUtils and evolutionUtils. It's an option to think about.

neurolib/optimize/exploration/exploration.py
jajcayn: I am afraid nothing... This can happen in `neurolib` native models, where some parameters can be arrays (e.g. `ext_x` in FHN brain net) and you wanna explore this parameter.. then, the respective coordinate for `DataArray` would have to be full array, which is not possible... 
User can do: 
* make sure all values in array are the same (then, this is saved with that one value - no problem)
* if she really needs heterogeneous external currents, use `MultiModel` where each node / mass has its own parameter, therefore this is not a problem

tests/test_exploration.py
jajcayn: I am setting attributes of xarray only for MultiModel, so this simply checks whether attributes are empty (empty `dict` evaluates to `False`)

jajcayn:
> I think most ideally there would be a `utils.xr` which collects all the xr related functions because they are starting to pop up in different parts of the code and I think that's not ideal. The same is true for pandas which is why I have been exporting functions to `explrationUtils.py` and `evolutionUtils.py`. Maybe that's even better than having a new file. It's an option to think about...

that'd be nice, but AFAIK we have two `xr()` functions at this point: one for Model superclass and one here... I was thinking before how or whether to try to squash them somehow into one, but I am afraid that these two are simply so different in the way **how** they construct the DataArray.. I cannot see how those two could be merged into one function... :/

caglorithm (commented):
> that'd be nice, but AFAIK we have two `xr()` functions at this point: one for Model superclass and one here... I was thinking before how or whether to try to squash them somehow into one, but I am afraid that these two are simply so different in the way **how** they construct the DataArray.. I cannot see how those two could be merged into one function... 😐

I think it's really nice to have it! We need a notebook on how to use it though... And probably it's time for a new version very soon!

PR #121, opened 2021-01-19
Support filtering from MultiModel parameters

jajcayn:
Star notation was always part of the MultiModel (getting all keys based on "*" in parameter dict). Now, filtering also works. Example:
```python
model.params["*tau"]
# returns
{'WCnode_0.WCmassEXC_0.tau': 2.5,
 'WCnode_0.WCmassEXC_0.noise_0.tau': 5.0,
 'WCnode_0.WCmassINH_1.tau': 3.75,
 'WCnode_0.WCmassINH_1.noise_0.tau': 5.0}
```
so we obtained all `tau` parameters - time scales for both masses in Wilson-Cowan model, but also time scales of Ornstein-Uhlenbeck noise... but:
```python
model.params["*tau|noise"]
# returns
{'WCnode_0.WCmassEXC_0.tau': 2.5, 'WCnode_0.WCmassINH_1.tau': 3.75}
```

Note: draft for now, I need to add some small things to `Evolution` due to this

### DIFF
--- a/neurolib/utils/collections.py
+++ b/neurolib/utils/collections.py
@@ -1,8 +1,11 @@
 """
 Collections of custom data structures and types.
 """
-from dpath.util import search, delete
+import random
+import string
 from collections import MutableMapping
+
+from dpath.util import delete, search
 
 DEFAULT_STAR_SEPARATOR = "."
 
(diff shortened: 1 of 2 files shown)


neurolib/utils/collections.py
```diff
@@ -26,9 +29,20 @@ def __setstate__(self, state):
 
 
 class star_dotdict(dotdict):
-    """Support star notation in dotdict. Nested dicts are now treated as glob"""
+    """
+    Supports star notation in dotdict. Nested dicts are now treated as glob.
+    Also supports minus as a pipe ("|") for filtering out strings after |.
+    """
```
caglorithm, line 35: Very nice stuff. Could you add one or two examples here?

caglorithm: The dotdict has become a pretty good object IMO :D

caglorithm (changes requested):

neurolib/utils/collections.py
caglorithm: I think we're pretty much good to go if you have time to add an example to the comments!

PR #128, opened 2021-01-21
Go wild with stimulation and/or noise

jajcayn:
Now you can concatenate model inputs like this:
```python
ou = OrnsteinUhlenbeckProcess(mu=0.1, sigma=0.02, tau=2.0, num_iid=2)
sq = SquareInput(amplitude=0.2, period=200.0, num_iid=2, stim_start=500)
sin = SinusoidalInput(amplitude=0.1, period=100, num_iid=2, stim_start=200)
step = StepInput(step_size=0.5, num_iid=2, stim_start=700)
conc = sq + sin + step + ou
plt.plot(conc.as_array(duration=1000, dt=0.1))
```
which gives you the sum of all of them... useful when you are stimulating + noise, or just wanna go full retard

image


caglorithm:
Mega nice! However, this should ideally be part of the core neurolib infrastructure and not only the multimodel part. Can this be implemented further below into utils?

jajcayn:
> Mega nice! However, this should ideally be part of the core neurolib infrastructure and not only the multimodel part. Can this be implemented further below into utils?

sure, but does it make sense? I mean the noise is integrated together with the full model inside one loop.. This little util pre-integrates / pre-generates noise+stimulation. Can the stimulus be pre-generated in the core neurolib?

caglorithm:
Yes, I mean this is a fantastic little tool for generating a time series and then putting it into the params of a model. for ALN it would be `model.params["ext_exc_current"] = conc.as_array(duration=1000, dt=0.1)`

jajcayn:
> Yes, I mean this is a fantastic little tool for generating a time series and then putting it into the params of a model. for ALN it would be `model.params["ext_exc_current"] = conc.as_array(duration=1000, dt=0.1)`.
> 
> Check the existing stimulus generation function in `functions.py`. I think this would be great to complement, or build un it. Ideally we can just use the same function. Or even better just create a stimulus.py file and put everything there maybe (including the old function). What do you think?

yes, totally makes sense! I'll move the files around!

jajcayn:
high line number just because of new files / removing old ones.. if you wanna check the actual changes, go per commits, just first two are actual changes, other three are just moving files around and changing imports

caglorithm (commented):
I wonder why this notebook didn't produce an error during checks...

https://github.com/neurolib-dev/neurolib/blob/master/examples/example-0.5-aln-external-stimulus.ipynb

Edit: because it was never tested... I forgot that we need to explicitly add every notebook ... somehow assumed that it tests all notebooks automagically.

caglorithm:
I've now took notebook tests completely out of the ci.yml and set up a new environment with treon to test the notebooks separately! :)

PR #152, opened 2021-04-14
Signal minor fix

jajcayn:
Quickie: `Signal` class minor fix and minor addition:
- [X] `iterate` now works in 1D arrays (by dummy adding a dimension)
- [X]  new function `rolling` - do your rolling mean, var, or whatever (the function is passed as an argument)


caglorithm:
<3

Issue #157, opened 2021-05-28
Speed up kuramato() function


vagechirkov:
Dear all,

I am planning to use `neurolib.utils.functions.kuromato` function to estimate Kuramoto order parameter but I faced the problem that running this function takes too long. 

Here is the time profiler report (it takes almost 10min to calculate Kuramoto order parameter for timeseries of 8 nodes and 20 seconds):

<details>

```bash

Timer unit: 1e-06 s

Total time: 588.338 s
/opt/anaconda3/envs/neurolib/lib/python3.7/site-packages/neurolib/utils/functions.py
Function: kuramoto at line 9

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     9                                           def kuramoto(traces, dt=0.1, smoothing=0.0, peakrange=[0.1, 0.2]):
    10                                               """
    11                                               Computes the Kuramoto order parameter of a timeseries which is a measure for synchrony.
    12                                               Can smooth timeseries if there is noise.
    13                                               Peaks are then detected using a peakfinder. From these peaks a phase is derived and then
    14                                               the amount of phase synchrony (the Kuramoto order parameter) is computed.
    15                                           
    16                                               :param traces: Multidimensional timeseries array
    17                                               :type traces: numpy.ndarray
    18                                               :param dt: Integration time step
    19                                               :type dt: float
    20                                               :param smoothing: Gaussian smoothing strength
    21                                               :type smoothing: float, optional
    22                                               :param peakrange: Width range of peaks for peak detection with `scipy.signal.find_peaks_cwt`
    23                                               :type peakrange: list[float], length 2
    24                                           
    25                                               :return: Timeseries of Kuramoto order paramter
    26                                               :rtype: numpy.ndarray
    27                                               """
    28         1         39.0     39.0      0.0      phases = []
    29         1          3.0      3.0      0.0      nTraces = len(traces)
    30         9         12.0      1.3      0.0      for n in range(nTraces):
    31         8     209572.0  26196.5      0.0          tList = np.dot(range(len(traces[n])), dt / 1000)
    32         8        102.0     12.8      0.0          a = traces[n]
    33                                           
    34                                                   # find peaks
    35         8         27.0      3.4      0.0          if smoothing > 0:
    36                                                       a = scipy.ndimage.filters.gaussian_filter(traces[n], smoothing)  # smooth data
[...]

PR #158, opened 2021-05-28
Speed up kuramoto function

vagechirkov:
This PR should significantly speed up the computation of Kuramoto order parameter.

Reference issue: #157

### DIFF
--- a/neurolib/utils/functions.py
+++ b/neurolib/utils/functions.py
@@ -26,7 +26,7 @@ def kuramoto(traces, smoothing=0.0, distance=10, prominence=5):
     :return: Timeseries of Kuramoto order paramter
     :rtype: numpy.ndarray
     """
-    @numba.njit(nopython=True)
+    @numba.njit
     def _estimate_phase(maximalist, n_times):
         lastMax = 0
         phases = np.empty((n_times), dtype=np.float64)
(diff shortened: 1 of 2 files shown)


neurolib/utils/functions.py
```diff
@@ -15,51 +16,63 @@ def kuramoto(traces, dt=0.1, smoothing=0.0, peakrange=[0.1, 0.2]):
...
+@jit(nopython=True)
+def _estimate_r(ntraces, times, phases):
+    kuramoto = np.empty((times), dtype=np.float64)
+    for t in range(times):
+        R = 1j*0
+        for n in range(ntraces):
+            R += np.exp(1j * phases[n, t])
+        R /= ntraces
+        kuramoto[t] = np.absolute(R)
     return kuramoto
```
caglorithm, line 76: Could you put these definitions inside the `kuramoto()` function please? Since they won't be used by any other function, it would be cleaner to do so. Hope that `numba.njit` doesn't have a problem with that! Otherwise, LGTM!

neurolib/utils/functions.py
```diff
@@ -1,12 +1,13 @@
 import logging
 import numpy as np
 import scipy.signal
+from numba import jit
```
caglorithm, line 4: Just for consistency: Please use the convention: 
`import numba` 
and decorate with 
`@numba.njit` instead of `@jit(nopython=True)`

caglorithm (changes requested):
First I thought we need `dt` but apparently we haven't really used it in the old version of `kuramoto` so it's fine that you got rid of it.  Please note the two inline comments I left. 

Afterwards, we can merge! 💪

caglorithm:
I'm guessing that the documentation build fails because it's a merge from an external repo? I'm getting a username-password error. Will ignore for now, should fix itself once it's merged to `master`

caglorithm:
Perfekt, LGTM, merging!

PR #174, opened 2021-07-07
Rename notebooks

caglorithm:
* Changed names of multimodel notebooks, will do an editing run later
* Fixed a bug in example-3 because of updated pandas


jajcayn:
making example 4.2 with significantly fewer runs (exploration and evolution) apparently helped, ready for merge

PR #178, opened 2021-08-04
BoxSearch with custom evaluate function

jajcayn:
quick bug fix:
now `MultiModel` exploration should work with custom evaluate function


caglorithm:
It's not much but it's honest work 👍🏽 merging

PR #184, opened 2021-09-03
QFix: Rescaling connectivity in ALN MultiModel

jajcayn:
quick fix in rescaling connectivity on parameter update in MultiModel's ALN


jajcayn:
Converting to a draft right now, we have found a bug with Ben, and it also affects this PR... will first fix the bug, then reopen

jajcayn:
Fixed in this branch in the end, since it's the same function... waiting for tests, then possible to merge

PR #206, opened 2022-05-13
time-dependent inputs FHN

lenasal:


neurolib/models/fhn/model.py
```diff
@@ -15,8 +15,8 @@ class FHNModel(Model):
     state_vars = ["x", "y", "x_ou", "y_ou"]
     output_vars = ["x", "y"]
     default_output = "x"
-    input_vars = ["x_ext", "y_ext"]
-    default_input = "x_ext"
```
caglorithm, line 19: These variables are for external stimulation and therefore they should be left as is.

neurolib/models/fhn/loadDefaultParams.py
```diff
@@ -78,6 +78,10 @@ def loadDefaultParams(Cmat=None, Dmat=None, seed=None):
     params.x_ext = np.ones((params.N,))
     params.y_ext = np.zeros((params.N,))
 
+    # time-dependent stimulation inputs
+    params.x_stim = 0.
+    params.y_stim = 0.
```
caglorithm, line 83: Instead of adding a new parameter, the old ones should be extended.

neurolib/models/fhn/timeIntegration.py
```diff
@@ -246,3 +255,46 @@ def timeIntegration_njit_elementwise(
...
+    if target.shape[1] > original.shape[1]:
+        rep_x = int(target.shape[1] / original.shape[1]) + 1
+    else:
+        rep_x = 1
+    original = np.tile(original, (1, rep_x))
+
+    # cut from end because the beginning can be initial condition
+    original = original[: target.shape[0], -target.shape[1] :]
+
+    return original
```
caglorithm, line 300: The existing `adjust_shape` function in `neurolib/models/aln/timeIntegration.py` needs to be refactored into either a new file in `neurolib/models/` or into `neurolib/utils/`.

I don't think there is an existing file yet where `adjust_shape` would fit.

caglorithm (changes requested):
Thanks for your first PR! 

Please see my comments in the code. Summary:

1) We don't need a new variable, the old ones should be simply extended
2) Please follow the template of how the parameter `ext_exc_current` of the `ALNModel` is defined and then adjusted. It allows for constant and time-varying inputs.
3) We need to refactor `adjust_shape` from the aln model so that the FHN model can use it as well (and all other models)
4) I would suggest adding time-dependent input to all models in one go.

Thank you!

neurolib/models/fhn/loadDefaultParams.py
lenasal: if we keep one set of external stimulation variables only, it is not entirely analogous to the ALN model, where we have "mue_ext_mean", "mui_ext_mean" as constant inputs that enable to determine the location in state-space, and "ext_exc_current", "ext_inh_current", which are time-dependent inputs. Also, there wouldn't be parameters to define the state-space location (which at least for the WC model might be problematic).

neurolib/models/fhn/timeIntegration.py
lenasal: I agree. Do you have a preference wrt `neurolb/models/` or `neurolib/utils/`? Also, do you have a suggestion for the name of the file?

jajcayn: personally, I'd do `neurolib/utils/model_utils.py`

PR #209, opened 2022-05-23
Time-dependent inputs

lenasal:
Changed parameters `x_ext `and `y_ext `in the FHN and the Hopf model to be time-dependent. Static inputs can be set using the parameters `x_ou_mean `and `y_ou_mean`.
- Changed minimal example of FHN: use `x_ou_mean` as bifurcation parameter
- moved `adjust_shape` function to seperate file `utils/adjust_params`


caglorithm:
Please remove all changes that are not related to this PR

neurolib/utils/adjust_params.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+
+def adjust_shape(original, target):
```
caglorithm, line 3: This should not go into a separate file. @jajcayn: What about `utils/collections.py`? We have only the `dotdict` type in there.

neurolib/models/hopf/timeIntegration.py
```diff
@@ -76,14 +77,14 @@ def timeIntegration(params):
...
-    y_ext = params["y_ext"]
-
     # state variable arrays, have length of t + startind
     # they store initial conditions AND simulated data
     xs = np.zeros((N, startind + len(t)))
     ys = np.zeros((N, startind + len(t)))
 
+    x_ext = ap.adjust_shape(params["x_ext"], xs)
+    y_ext = ap.adjust_shape(params["y_ext"], ys)
+
```
caglorithm, line 87: This should not be necessary! Why is this duplicated? It should be done in `loadDefaultParams.py` only?

neurolib/models/fhn/loadDefaultParams.py
```diff
@@ -74,9 +74,9 @@ def loadDefaultParams(Cmat=None, Dmat=None, seed=None):
     params.x_ou = np.zeros((params.N,))
     params.y_ou = np.zeros((params.N,))
 
-    # values of the external inputs
-    params.x_ext = np.ones((params.N,))
-    params.y_ext = np.zeros((params.N,))
+    # values of external inputs
+    params.x_ext = 0.
+    params.y_ext = 0.
```
caglorithm, line 79: Do not change this.

caglorithm:
All tests are still failing. Can you check?

PR #210, opened 2022-06-03
Minor fixes: setting seed + avoiding index out of bounds for noise array

mattiasJohnson:
Two minor things I encountered when working on my thalamus-aln implementation.

**Seed not set**
Seed given as argument was not set if the constructor called `loadDefaultParams()`
```python
# Before fix:
model = ThalamicMassModel(seed=42)
print(model.params["seed"])  # --> None
```

**Index out of bounds in `noise`**
`noise` has the size `len(t)` whilst `i` goes over `range(startind, startind + len(t))`.

### DIFF
--- a/neurolib/models/thalamus/model.py
+++ b/neurolib/models/thalamus/model.py
@@ -6,7 +6,7 @@ from . import timeIntegration as ti
 class ThalamicMassModel(Model):
     """
     Two population thalamic model
-    
+
     Reference:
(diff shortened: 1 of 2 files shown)


caglorithm:
Thank you for this PR and for finding these issues. Generally speaking, even small changes like that should be split up into different PR's but for this one we can keep it in. Running tests and seeing if everything works, if so, I'll merge!

PR #211, opened 2022-06-08
Time dependent inputs for FHN, WC, Hopf

lenasal:
Include time-dependent inputs to the FHN model, the WC model, and the Hopf model.
- move "adjust_shape" function from aln time integration file to general file in utils
- fix bug in test_models
- add tests for stimulation of FHN, WC, Hopf


caglorithm:
* Tests still failing. 
* PR contains changes in notebooks. 
* Changes in parameter names need to be reverted everywhere (also in tests).

lenasal:
I've updated things, test are not failing anymore.

lenasal:
changed index of inputs in time integration, this way it is more in line with the aln model (and also, I personally think it makes more sense)

caglorithm:
Nice! Could you also remove your commits for the notebooks from the PR?

neurolib/models/hopf/loadDefaultParams.py
```diff
@@ -77,8 +77,8 @@ def loadDefaultParams(Cmat=None, Dmat=None, seed=None):
     params.y_ou = np.zeros((params.N,))
 
     # values of the external inputs
-    params.x_ext = np.zeros((params.N,))
-    params.y_ext = np.zeros((params.N,))
```
caglorithm, line 81: Leaving these lines the way they are should work as well, right?

caglorithm (changes requested):
Please see the one nit comment I've made!

caglorithm:
LGTM. Merging after tests finish ✅.

PR #222, opened 2022-09-20
Seed default parameter

lenasal:
Corrected the typo in the loadDefaultParameters.py file of the Hopf model. Previously, seed was set to 0 by default, which made the random numbers predictable, and resulted in the same simulation results repeatedly.


jajcayn (approved):
LGTM :)

jajcayn:
@lenasal please correct failing tests
thanks:)

PR #228, opened 2022-11-02
Fix tests

jajcayn:
The title says it all.
Anyway, @caglorithm, we should discuss maybe dropping support for lower python than 3.8.. The thing is, `xarray` did drop the support for 3.7 (and obviously for 3.6) some time ago (https://github.com/pydata/xarray/issues/7149#issuecomment-1272537456), so now loading .nc files (in `Signal` class, as done in example-3) requires hacky workaround - we need to downgrade `importlib-metadata` (dependency of `xarray`)...
The problem, of course, is that due to `pypet`, only python 3.6 and 3.7 work on macOS... so dropping support for 3.6 and 3.7 would mean dropping support for macOS :/ Or, getting rid of `pypet`...
But for now, tests should work...

@lenasal after this is merged, please rebase your PR onto the new master; then it should work for you:)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies 🛠
         run: |
-          python -m pip install --upgrade pip
-          pip install flake8 codecov pytest-cov wheel setuptools matplotlib seaborn mne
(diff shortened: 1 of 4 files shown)


caglorithm:
MVP! Thank you @jajcayn very much needed :)

caglorithm:
I agree, let's drop support for anything < 3.8

PR #230, opened 2022-12-07
Fix spelling Example 0.0

StevenGeysen:
While going through the examples, I noticed some spelling mistakes and fixed them.
I hope it is okay to do it like this.

Thanks for the work and sharing the framework


caglorithm:
Thank you for this PR!

PR #232, opened 2023-04-24
fixed bug in fhn, hopf, wc, ww models: copy params

lenasal:
x_ou, y_ou and exc_ou, inh_ou variables overwrite corresponding params, include ".copy()" to fix this


caglorithm (approved):
LGTM

PR #233, opened 2023-04-24
keep e, i in range [0,1]

lenasal:
e,i variables can fall below 0 or exceed 1, which is unphysical


neurolib/models/wc/timeIntegration.py
```diff
@@ -243,6 +243,13 @@ def S_I(x):
             excs[no, i] = excs[no, i - 1] + dt * exc_rhs
             inhs[no, i] = inhs[no, i - 1] + dt * inh_rhs
 
+            # make sure e and i variables do not exceed 1 (can only happen with noise)
+            excs[no, i] = min(1.0, excs[no, i])
+            inhs[no, i] = min(1.0, inhs[no, i])
+            # make sure e and i variables do decrease below 0 (can only happen with noise)
+            excs[no, i] = max(0.0, excs[no, i])
+            inhs[no, i] = max(0.0, inhs[no, i])
```
caglorithm, line 251: This is too costly IMO.

This is how we do it in `aln`

```python
            if seev[no] < 0:
                seev[no] = 0.0

            if siev[no] < 0:
                siev[no] = 0.0

            if seiv[no] < 0:
                seiv[no] = 0.0

            if siiv[no] < 0:
                siiv[no] = 0.0
```

jajcayn: agreed, I'd use `if` -> set

callebtc (approved):
LGTM!

PR #234, opened 2023-04-24
removed Dmat_ndt as param set in timeIntegration file

lenasal:
This PR removes `params["Dmat_ndt"] = Dmat_ndt` as a parameter from the FHN, Hopf, WC, WW models because `Dmat_ndt` is derived from `lengthMat` and is not a parameter that should be set by the user.


caglorithm (approved):
LGTM

jajcayn (approved):
LGTM:)

PR #239, opened 2023-07-26
Kuramoto dev

bramantyois:
* Kuramoto model implementation at ./neurolib/models/kuramoto/*
* Example notebook ./examples/example-0.5-kuramoto.ipynb
* Other example notebooks > 0.5 are shifted


lenasal:
modulo is supported by numba (see https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html), please include in time integration

bramantyois:
added mod

PR #255, opened 2024-01-05
Fix bug in fhn and hopf OC

lenasal:
network jacobian was not computed correctly for diffusive coupling in fhn and hopf model


lenasal:
closing pull request was by accident, sorry!

PR #260, opened 2024-01-29
Bug Fix: Setting value of tau_inh in timeIntegration

Andrew-Clappison:
Hello, I observed a discrepancy in analysis using models.ww when compared to analysis from another implementation. Further experimentation led me to notice the steady state values were different from the analytical solution (in the non-stochastic case). Luckily, I was able to identify the source of the issue, which was simply setting tau_inh incorrectly in models.ww.timeIntegration. Hopefully this can be merged quickly. Thank you!


lenasal:
Nice catch, thank you!

Issue #254, opened 2023-12-14
Example files appear to be broken


JacobDuda:
The example-0-aln-minimal notebook is encountering some kind of error, even when being run through the binder found on the GitHub page. The final plots are not returning the results that the documentation suggests that they should, I am not sure on any possible causes of this to be honest, just wanted to point this out so the authors are aware. Thanks
<img width="419" alt="neurolib_example_error_1" src="https://github.com/neurolib-dev/neurolib/assets/78237315/8220cadc-0d97-4473-ad07-30983d563ea1">
<img width="424" alt="neurolib_example_error_2" src="https://github.com/neurolib-dev/neurolib/assets/78237315/80ab1e28-88dd-4582-9e12-7fc354386c27">

Issue #250, opened 2023-09-28
Parameters provided in examples (and default parameters) does not perform BOLD simulation correctly


yusenz:
For some reason I do not understand, the append_outputs parameter in `model.run()` is defaulted to `None` and thus `False` in `model.integrateChunkwise()`. Meanwhile, the bold simulation in `model.simulateBold()` pulls from `self.default_output`.
As a result, the BOLD simulation, when running with default parameters, will ONLY use the simulated activity of last chunk to simulate the entire duration. This is especially noticable when the model is run with time-dependent external stimuli.

The examples with BOLD like example 0 aln model suffer from this issue, and it has a strong misleading effect.

yusenz:
Edit: setting `append_outputs=True` does not fix the issue. The bold activity still behaves like the stimuli is not time dependent at all. The previous comment on the cause of the issue might be wrong.

PR #261, opened 2024-03-05
spring cleaning: removing bugs and code smells

1b15:
- addresses https://github.com/neurolib-dev/neurolib/issues/250 and https://github.com/neurolib-dev/neurolib/issues/254
- fixes inconsistencies and unexpected behaviour in the implementation and interaction of chunking, continuing runs, bold, and appending outputs
- improves maintainability by removing various code smells (e.g., low cohesion because of global bold parameter, repeated implementation of the same output saving code, etc)

### DIFF
--- a/neurolib/models/model.py
+++ b/neurolib/models/model.py
@@ -324,9 +324,6 @@ class Model:
     def setInitialValuesToLastState(self):
         """Reads the last state of the model and sets the initial conditions to that state for continuing a simulation."""
         if not all([sv in self.state for sv in self.state_vars]):
-            logging.warning(
-                f"`setInitialValuesToLastState` was called, but the state dict is incomplete. This is probably caused by `continue_run=True` on the first model run."
-            )
             return
         for iv, sv in zip(self.init_vars, self.state_vars):
             # if state variables are one-dimensional (in space only)
--- a/neurolib/models/multimodel/model.py
+++ b/neurolib/models/multimodel/model.py
@@ -203,9 +203,6 @@ class MultiModel(Model):
 
     def setInitialValuesToLastState(self):
         if not self.state:
-            logging.warning(
-                f"`setInitialValuesToLastState` was called, but the state dict is empty. This is probably caused by `continue_run=True` on the first model run."
-            )
             return
         new_initial_state = np.zeros((self.model_instance.initial_state.shape[0], self.maxDelay + 1))
         total_vars_counter = 0


neurolib/models/bold/model.py
```diff
@@ -67,19 +66,8 @@ def run(self, activity, append=False):
...
-            self.BOLD = BOLD_resampled
-        elif append is True:
-            # append new data to old data
-            self.t_BOLD = np.hstack((self.t_BOLD, t_BOLD_resampled))
-            self.BOLD = np.hstack((self.BOLD, BOLD_resampled))
-        else:
-            # overwrite old data
-            self.t_BOLD = t_BOLD_resampled
-            self.BOLD = BOLD_resampled
-
```
caglorithm, line 82: You seem to have the `append=True` case, why?

neurolib/models/model.py
```diff
@@ -311,7 +299,7 @@ def clearModelState(self):
         self.state = dotdict({})
         self.outputs = dotdict({})
         # reinitialize bold model
-        if self.params.get("bold"):
+        if self.boldInitialized:
```
caglorithm, line 302: Does this do the same thing as before? Reads like: if it is initialized, initialize it.

neurolib/models/model.py
```diff
@@ -335,6 +323,8 @@ def storeOutputsAndStates(self, t, variables, append=False):
 
     def setInitialValuesToLastState(self):
         """Reads the last state of the model and sets the initial conditions to that state for continuing a simulation."""
+        if not hasattr(self, "t"):
+            raise ValueError("You tried using continue_run=True on the first run.")
```
caglorithm, line 327: Why do we have to error here? The user could user `continue_run=True` in the first call as well.

neurolib/models/model.py
```diff
@@ -474,25 +464,28 @@ def setOutput(self, name, data, append=False, removeICs=False):
                 raise ValueError(f"Don't know how to truncate data of shape {data.shape}.")
 
         # subsample to sampling dt
-        if data.ndim == 1:
-            data = data[:: self.sample_every]
-        elif data.ndim == 2:
-            data = data[:, :: self.sample_every]
-        else:
-            raise ValueError(f"Don't know how to subsample data of shape {data.shape}.")
+        if data.shape[-1] >= self.params["duration"] - self.startindt:
```
caglorithm, line 467: Does this `if` clause prevent any previous case to run?

neurolib/models/multimodel/model.py
```diff
@@ -128,27 +126,22 @@ def run(
...
 
-        # TODO: legacy argument support
-        if append_outputs is not None:
-            append = append_outputs
-
         # if a previous run is not to be continued clear the model's state
-        if continue_run is False:
+        if continue_run:
+            self.setInitialValuesToLastState()
+        else:
```
caglorithm, line 138: We would like to allow `continue_run` even in the case if there hasn't been any previous run.

neurolib/models/multimodel/model.py
```diff
@@ -236,44 +228,10 @@ def integrateChunkwise(self, chunksize, bold, append_outputs):
 
     def storeOutputsAndStates(self, results, append):
         # save time array
-        self.setOutput("t", results.time.values + self.start_t, append=append, removeICs=False)
-        self.setStateVariables("t", results.time.values + self.start_t)
+        self.setOutput("t", results.time.values, append=append, removeICs=False)
+        self.setStateVariables("t", results.time.values)
```
caglorithm, line 232: This changes the time axis.

caglorithm (changes requested):
Thank you very much for this awesome PR. Please have a look at the comments I left. 

The only thing I can see is that `continue_run` might not behave as expected and that there is a change of the time axis in `storeOutputsAndStates`.

neurolib/models/bold/model.py
1b15: The functionality for saving outputs is already in `models/model.py`.

neurolib/models/model.py
1b15: The difference is that BOLD is now only initialized at the beginning of the first run. One only needs to clear the bold state with re-initialization if it has been initialized before.

neurolib/models/model.py
1b15: This catches unintended subsampling of BOLD.

neurolib/models/multimodel/model.py
1b15: We previously had different time axis behaviours for model and multimodel:
- for regular models, the time axis always started at 0 because it was reset for every chunk unless it was appended
- for multimodels, the time axis started at `self.start_t` from continued runs

We adapted the multimodel behaviour with a more elegant implementation without `self.start_t`.

neurolib/models/model.py
1b15: addressed this in the latest commit

neurolib/models/multimodel/model.py
```diff
@@ -205,17 +196,14 @@ def integrate(self, append_outputs=False, simulate_bold=False, noise_input=None)
...
 
         # bold simulation after integration
         if simulate_bold and self.boldInitialized:
             self.simulateBold(result[self.default_output].values.T, append=True)
 
     def setInitialValuesToLastState(self):
-        # set start t for next run for the last value now
-        self.start_t = self.t[-1]
+        if not hasattr(self, "t"):
+            raise ValueError("You tried using continue_run=True on the first run.")
```
VascoSch92, line 206:
```suggestion
            raise ValueError("You tried using `continue_run=True` on the first run.")
```

1b15: thank you Vasco, I forgot to make the suggested change in the multimodel

caglorithm (approved):
LGTM, amazing work, thank you

